PR #33, opened 2025-04-25
Update pauser to handle exception and run unpause as long as possible

kota2and3kan:
## Description

This PR added error handling to the `Puaser.pause()` method.

Previously, there is no error handling, and it might cause `Scalar products keep pause state forever` when some error occurs.

In other words, it might cause some issues like the system down.

To address such issue, I added error handling. And now, if some error occurs, the `Pauser` tries to run `unpause()` method several times to make Scalar products the unpause state as long as possible.

Also, I updated it to throw the `PauseException`. So, the caller (CLI or some other tools that implement this library) can notice the above issue and handle it on their side.

Please take a look!

## Related issues and/or PRs

N/A

## Changes made (By copilot)

This pull request introduces improvements to the `Pauser` class in `lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java` to enhance error handling, testability, and maintainability. It also includes a minor addition to the project's dependencies in `lib/build.gradle`. The most important changes are grouped below by theme.

### Enhancements to error handling:
* Improved error handling in the `pause` method by adding nested `try-catch` blocks to ensure that the `unpauseWithRetry` method is called even if the `pause` operation fails. This ensures a fallback mechanism for unpausing pods and provides detailed error messages for critical failures.
* Updated the `unpauseWithRetry` method to throw a `PauserException` with a detailed error message when retry attempts fail. This ensures that users are informed of the critical issue and the steps required to resolve it.

### Testability improvements:
* Annotated the `MAX_UNPAUSE_RETRY_COUNT`, `unpauseWithRetry`, `getTarget`, and `getRequestCoordinator` members with `@VisibleForTesting` to make them accessible for unit testing. This improves the testability of the `Pauser` class. [[1]](diffhunk://#diff-529e2e818d31d239079b031928623cab070acd5743e317b3d709c34e6e52b394L37-R38) [[2]](diffhunk://#diff-529e2e818d31d239079b031928623cab070acd5743e317b3d709c34e6e52b394L116-R143) [[3]](diffhunk://#diff-529e2e818d31d239079b031928623cab070acd5743e317b3d709c34e6e52b394L126-R179)

### Dependency updates:
* Added the `mockito-inline` library to the `testImplementation` dependencies in `lib/build.gradle` to support mocking of static methods and final classes in tests.

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation to reflect the changes.
- [x] Any remaining open issues linked to this PR are documented and up-to-date (Jira, GitHub, etc.).
- [x] Tests (unit, integration, etc.) have been added for the changes.
- [x] My changes generate no new warnings.
- [x] Any dependent changes in other PRs have been merged and published.

## Additional notes (optional)

N/A

### DIFF
--- a/lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
+++ b/lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
@@ -247,7 +247,8 @@ public class Pauser {
       pauserException = unpauseFailedException;
     }
 
-    // Treat the pause failure as the second priority because the issue might be caused by a configuration mistake.
+    // Treat the pause failure as the second priority because the issue might be caused by a
+    // configuration mistake.
     if (pauseFailedException != null) {
       if (pauserException == null) {
         pauserException = pauseFailedException;
@@ -256,8 +257,8 @@ public class Pauser {
       }
     }
 
-    // Treat the getting target failure as the third priority because this issue might be caused by a temporary glitch, for
-    // example, network failures.
+    // Treat the getting target failure as the third priority because this issue might be caused by
+    // a temporary glitch, for example, network failures.
     if (getTargetAfterPauseFailedException != null) {
       if (pauserException == null) {
         pauserException = getTargetAfterPauseFailedException;
@@ -266,8 +267,9 @@ public class Pauser {
       }
     }
 
-    // Treat the status checking failure as third priority because this issue might be caused by a temporary glitch,
-    // for example, targetAfterPause is null.
+    // Treat the status checking failure as the third priority because this issue might be caused by
+    // a temporary glitch, for example, getting target information by using the Kubernetes API fails
+    // after the pause operation.
     if (statusCheckFailedException != null) {
       if (pauserException == null) {
         pauserException = statusCheckFailedException;
@@ -276,8 +278,8 @@ public class Pauser {
       }
     }
 
-    // Treat the status unmatched issue as third priority because this issue might be caused by temporary
-    // glitch, for example, pod restarts.
+    // Treat the status unmatched issue as the third priority because this issue might be caused by
+    // temporary glitch, for example, pod restarts.
     if (statusUnmatchedException != null) {
       if (pauserException == null) {
         pauserException = statusUnmatchedException;
--- a/lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
+++ b/lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
@@ -130,7 +130,7 @@ public class Pauser {
     String statusDifferentErrorMessage =
         "The target pods were updated during the pause duration. You cannot use the backup that"
             + " was taken during this pause duration. ";
-    String unpauseFailedButBackupOkErrorMessage =
+    String unpauseFailedButPauseOkErrorMessage =
         String.format(
             "Note that the pause operations for taking backup succeeded. You can use a backup that"
                 + " was taken during this pause duration: Start Time = %s, End Time = %s. ",
@@ -166,14 +166,14 @@ public class Pauser {
 
     // If both the pause operation and status check succeeded, you can use the backup that was taken
     // during the pause duration.
-    boolean backupOk = (pauseFailedException == null) && compareTargetSuccessful;
+    boolean isPauseOk = (pauseFailedException == null) && compareTargetSuccessful;
 
     // Create an error message if any of the operations failed.
     StringBuilder errorMessageBuilder = new StringBuilder();
     if (unpauseFailedException != null) {
       errorMessageBuilder.append(unpauseErrorMessage);
-      if (backupOk) {
-        errorMessageBuilder.append(unpauseFailedButBackupOkErrorMessage);
+      if (isPauseOk) {
+        errorMessageBuilder.append(unpauseFailedButPauseOkErrorMessage);
       }
     }
     if (pauseFailedException != null) {
--- a/lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
+++ b/lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
@@ -37,24 +37,33 @@ public class Pauser {
 
   private final TargetSelector targetSelector;
 
-  private static final String UNPAUSE_ERROR_MESSAGE =
+  @VisibleForTesting
+  static final String UNPAUSE_ERROR_MESSAGE =
       "Unpause operation failed. Scalar products might still be in a paused state. You"
           + " must restart related pods by using the `kubectl rollout restart deployment"
-          + " <DEPLOYMENT_NAME>` command to unpause all pods. ";
-  private static final String PAUSE_ERROR_MESSAGE =
+          + " <DEPLOYMENT_NAME>` command to unpause all pods.";
+
+  @VisibleForTesting
+  static final String PAUSE_ERROR_MESSAGE =
       "Pause operation failed. You cannot use the backup that was taken during this pause"
           + " duration. You need to retry the pause operation from the beginning to"
-          + " take a backup. ";
-  private static final String GET_TARGET_AFTER_PAUSE_ERROR_MESSAGE =
+          + " take a backup.";
+
+  @VisibleForTesting
+  static final String GET_TARGET_AFTER_PAUSE_ERROR_MESSAGE =
       "Failed to find the target pods to examine if the targets pods were updated during"
-          + " paused. ";
-  private static final String STATUS_CHECK_ERROR_MESSAGE =
+          + " paused.";
+
+  @VisibleForTesting
+  static final String STATUS_CHECK_ERROR_MESSAGE =
       "Status check failed. You cannot use the backup that was taken during this pause"
           + " duration. You need to retry the pause operation from the beginning to"
-          + " take a backup. ";
-  private static final String STATUS_DIFFERENT_ERROR_MESSAGE =
+          + " take a backup.";
+
+  @VisibleForTesting
+  static final String STATUS_UNMATCHED_ERROR_MESSAGE =
       "The target pods were updated during the pause duration. You cannot use the backup that"
-          + " was taken during this pause duration. ";
+          + " was taken during this pause duration.";
 
   /**
    * @param namespace The namespace where the pods are deployed.
@@ -120,7 +129,7 @@ public class Pauser {
     try {
       pausedDuration = pauseInternal(requestCoordinator, pauseDuration, maxPauseWaitTime);
     } catch (Exception e) {
-      pauseFailedException = new PauseFailedException("Pause operation failed.", e);
+      pauseFailedException = new PauseFailedException(PAUSE_ERROR_MESSAGE, e);
     }
 
     // Run an unpause operation.
@@ -128,58 +137,45 @@ public class Pauser {
     try {
       unpauseWithRetry(requestCoordinator, MAX_UNPAUSE_RETRY_COUNT);
     } catch (Exception e) {
-      unpauseFailedException = new UnpauseFailedException("Unpause operation failed.", e);
+      unpauseFailedException = new UnpauseFailedException(UNPAUSE_ERROR_MESSAGE, e);
     }
 
     // Get pods and deployment information after pause.
-    TargetSnapshot targetAfterPause;
+    TargetSnapshot targetAfterPause = null;
+    GetTargetAfterPauseFailedException getTargetAfterPauseFailedException = null;
     try {
       targetAfterPause = getTarget();
     } catch (Exception e) {
-      PauserException pauserException =
-          new PauserException(GET_TARGET_AFTER_PAUSE_ERROR_MESSAGE, e);
-      if (unpauseFailedException == null) {
-        throw pauserException;
-      } else {
-        throw new UnpauseFailedException(UNPAUSE_ERROR_MESSAGE, pauserException);
-      }
+      getTargetAfterPauseFailedException =
+          new GetTargetAfterPauseFailedException(GET_TARGET_AFTER_PAUSE_ERROR_MESSAGE, e);
     }
 
     // Check if pods and deployment information are the same between before pause and after pause.
-    boolean isTargetStatusEqual;
+    StatusCheckFailedException statusCheckFailedException = null;
+    StatusUnmatchedException statusUnmatchedException = null;
     try {
-      isTargetStatusEqual = targetStatusEquals(targetBeforePause, targetAfterPause);
+      statusUnmatchedException = targetStatusEquals(targetBeforePause, targetAfterPause);
     } catch (Exception e) {
-      StatusCheckFailedException statusCheckFailedException =
-          new StatusCheckFailedException(STATUS_CHECK_ERROR_MESSAGE, e);
-      if (unpauseFailedException == null) {
-        throw statusCheckFailedException;
-      } else {
-        throw new UnpauseFailedException(UNPAUSE_ERROR_MESSAGE, statusCheckFailedException);
-      }
+      statusCheckFailedException = new StatusCheckFailedException(STATUS_CHECK_ERROR_MESSAGE, e);
     }
-
-    // Create an error message if any of the operations failed.
-    String errorMessage =
-        createErrorMessage(unpauseFailedException, pauseFailedException, isTargetStatusEqual);
 
     // We use the exceptions as conditions instead of using boolean flags like `isPauseOk`, etc. If
     // we use boolean flags, it might cause a bit large number of combinations. For example, if we
     // have three flags, they generate 2^3 = 8 combinations. It also might make the nested if
     // statement or a lot of branches of the switch statement. That's why we don't use status flags
     // for now.
+    PauserException pauserException =
+        buildException(
+            unpauseFailedException,
+            pauseFailedException,
+            getTargetAfterPauseFailedException,
+            statusCheckFailedException,
+            statusUnmatchedException);
 
     // Return the final result based on each process.
-    if (unpauseFailedException != null) { // Unpause Failed.
-      // Unpause issue is the most critical because it might cause system down.
-      throw new UnpauseFailedException(errorMessage, unpauseFailedException);
-    } else if (pauseFailedException != null) { // Pause Failed.
-      // Pause Failed is second priority because pause issue might be caused by configuration error.
-      throw new PauseFailedException(errorMessage, pauseFailedException);
-    } else if (!isTargetStatusEqual) { // Target status changed.
-      // Target status changed is third priority because this issue might be caused by temporary
-      // issue, for example, pod restarts.
-      throw new PauseFailedException(errorMessage);
+    if (pauserException != null) {
+      // Some operations failed.
+      throw pauserException;
     } else {
       // All operations succeeded.
       return pausedDuration;
@@ -225,24 +221,70 @@ public class Pauser {
   }
 
   @VisibleForTesting
-  boolean targetStatusEquals(TargetSnapshot before, TargetSnapshot after) {
-    return before.getStatus().equals(after.getStatus());
+  @Nullable
+  StatusUnmatchedException targetStatusEquals(
+      TargetSnapshot before, @Nullable TargetSnapshot after) {
+    if (after != null && before.getStatus().equals(after.getStatus())) {
+      return null;
+    } else {
+      return new StatusUnmatchedException(STATUS_UNMATCHED_ERROR_MESSAGE);
+    }
   }
 
-  private String createErrorMessage(
-      UnpauseFailedException unpauseFailedException,
-      PauseFailedException pauseFailedException,
-      boolean isTargetStatusEqual) {
-    StringBuilder errorMessageBuilder = new StringBuilder();
+  @Nullable
+  @VisibleForTesting
+  PauserException buildException(
+      @Nullable UnpauseFailedException unpauseFailedException,
+      @Nullable PauseFailedException pauseFailedException,
+      @Nullable GetTargetAfterPauseFailedException getTargetAfterPauseFailedException,
+      @Nullable StatusCheckFailedException statusCheckFailedException,
+      @Nullable StatusUnmatchedException statusUnmatchedException) {
+    PauserException pauserException = null;
+
+    // Unpause issue is the most critical because it might cause system down.
     if (unpauseFailedException != null) {
-      errorMessageBuilder.append(UNPAUSE_ERROR_MESSAGE);
+      pauserException = unpauseFailedException;
     }
+
+    // Pause Failed is second priority because pause issue might be caused by configuration error.
     if (pauseFailedException != null) {
-      errorMessageBuilder.append(PAUSE_ERROR_MESSAGE);
+      if (pauserException == null) {
+        pauserException = pauseFailedException;
+      } else {
+        pauserException.addSuppressed(pauseFailedException);
+      }
     }
-    if (!isTargetStatusEqual) {
-      errorMessageBuilder.append(STATUS_DIFFERENT_ERROR_MESSAGE);
+
+    // Get target issue is third priority because this issue might be caused by temporary issue, for
+    // example, network issues.
+    if (getTargetAfterPauseFailedException != null) {
+      if (pauserException == null) {
+        pauserException = getTargetAfterPauseFailedException;
+      } else {
+        pauserException.addSuppressed(getTargetAfterPauseFailedException);
+      }
     }
-    return errorMessageBuilder.toString();
+
+    // Status check issue is third priority because this issue might be caused by temporary issue,
+    // for example, targetAfterPause is null.
+    if (statusCheckFailedException != null) {
+      if (pauserException == null) {
+        pauserException = statusCheckFailedException;
+      } else {
+        pauserException.addSuppressed(statusCheckFailedException);
+      }
+    }
+
+    // Status unmatched issue is third priority because this issue might be caused by temporary
+    // issue, for example, pod restarts.
+    if (statusUnmatchedException != null) {
+      if (pauserException == null) {
+        pauserException = statusUnmatchedException;
+      } else {
+        pauserException.addSuppressed(statusUnmatchedException);
+      }
+    }
+
+    return pauserException;
   }
 }
--- a/lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
+++ b/lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
@@ -37,6 +37,25 @@ public class Pauser {
 
   private final TargetSelector targetSelector;
 
+  private static final String UNPAUSE_ERROR_MESSAGE =
+      "Unpause operation failed. Scalar products might still be in a paused state. You"
+          + " must restart related pods by using the `kubectl rollout restart deployment"
+          + " <DEPLOYMENT_NAME>` command to unpause all pods. ";
+  private static final String PAUSE_ERROR_MESSAGE =
+      "Pause operation failed. You cannot use the backup that was taken during this pause"
+          + " duration. You need to retry the pause operation from the beginning to"
+          + " take a backup. ";
+  private static final String GET_TARGET_AFTER_PAUSE_ERROR_MESSAGE =
+      "Failed to find the target pods to examine if the targets pods were updated during"
+          + " paused. ";
+  private static final String STATUS_CHECK_ERROR_MESSAGE =
+      "Status check failed. You cannot use the backup that was taken during this pause"
+          + " duration. You need to retry the pause operation from the beginning to"
+          + " take a backup. ";
+  private static final String STATUS_DIFFERENT_ERROR_MESSAGE =
+      "The target pods were updated during the pause duration. You cannot use the backup that"
+          + " was taken during this pause duration. ";
+
   /**
    * @param namespace The namespace where the pods are deployed.
    * @param helmReleaseName The Helm release name used to deploy the pods.
@@ -108,38 +127,17 @@ public class Pauser {
       unpauseFailedException = new UnpauseFailedException("Unpause operation failed.", e);
     }
 
-    // Prepare error messages for each process.
-    String unpauseErrorMessage =
-        String.format(
-            "Unpause operation failed. Scalar products might still be in a paused state. You"
-                + " must restart related pods by using the `kubectl rollout restart deployment"
-                + " %s` command to unpause all pods. ",
-            targetBeforePause.getDeployment().getMetadata().getName());
-    String pauseErrorMessage =
-        "Pause operation failed. You cannot use the backup that was taken during this pause"
-            + " duration. You need to retry the pause operation from the beginning to"
-            + " take a backup. ";
-    String getTargetAfterPauseErrorMessage =
-        "Failed to find the target pods to examine if the targets pods were updated during"
-            + " paused. ";
-    String statusCheckErrorMessage =
-        "Status check failed. You cannot use the backup that was taken during this pause"
-            + " duration. You need to retry the pause operation from the beginning to"
-            + " take a backup. ";
-    String statusDifferentErrorMessage =
-        "The target pods were updated during the pause duration. You cannot use the backup that"
-            + " was taken during this pause duration. ";
-
     // Get pods and deployment information after pause.
     TargetSnapshot targetAfterPause;
     try {
       targetAfterPause = getTarget();
     } catch (Exception e) {
-      PauserException pauserException = new PauserException(getTargetAfterPauseErrorMessage, e);
+      PauserException pauserException =
+          new PauserException(GET_TARGET_AFTER_PAUSE_ERROR_MESSAGE, e);
       if (unpauseFailedException == null) {
         throw pauserException;
       } else {
-        throw new UnpauseFailedException(unpauseErrorMessage, pauserException);
+        throw new UnpauseFailedException(UNPAUSE_ERROR_MESSAGE, pauserException);
       }
     }
 
@@ -149,26 +147,17 @@ public class Pauser {
       isTargetStatusEqual = targetStatusEquals(targetBeforePause, targetAfterPause);
     } catch (Exception e) {
       StatusCheckFailedException statusCheckFailedException =
-          new StatusCheckFailedException(statusCheckErrorMessage, e);
+          new StatusCheckFailedException(STATUS_CHECK_ERROR_MESSAGE, e);
       if (unpauseFailedException == null) {
         throw statusCheckFailedException;
       } else {
-        throw new UnpauseFailedException(unpauseErrorMessage, statusCheckFailedException);
+        throw new UnpauseFailedException(UNPAUSE_ERROR_MESSAGE, statusCheckFailedException);
       }
     }
 
     // Create an error message if any of the operations failed.
-    StringBuilder errorMessageBuilder = new StringBuilder();
-    if (unpauseFailedException != null) {
-      errorMessageBuilder.append(unpauseErrorMessage);
-    }
-    if (pauseFailedException != null) {
-      errorMessageBuilder.append(pauseErrorMessage);
-    }
-    if (!isTargetStatusEqual) {
-      errorMessageBuilder.append(statusDifferentErrorMessage);
-    }
-    String errorMessage = errorMessageBuilder.toString();
+    String errorMessage =
+        createErrorMessage(unpauseFailedException, pauseFailedException, isTargetStatusEqual);
 
     // Return the final result based on each process.
     if (unpauseFailedException != null) { // Unpause Failed.
@@ -228,5 +217,22 @@ public class Pauser {
   @VisibleForTesting
   boolean targetStatusEquals(TargetSnapshot before, TargetSnapshot after) {
     return before.getStatus().equals(after.getStatus());
+  }
+
+  private String createErrorMessage(
+      UnpauseFailedException unpauseFailedException,
+      PauseFailedException pauseFailedException,
+      boolean isTargetStatusEqual) {
+    StringBuilder errorMessageBuilder = new StringBuilder();
+    if (unpauseFailedException != null) {
+      errorMessageBuilder.append(UNPAUSE_ERROR_MESSAGE);
+    }
+    if (pauseFailedException != null) {
+      errorMessageBuilder.append(PAUSE_ERROR_MESSAGE);
+    }
+    if (!isTargetStatusEqual) {
+      errorMessageBuilder.append(STATUS_DIFFERENT_ERROR_MESSAGE);
+    }
+    return errorMessageBuilder.toString();
   }
 }
--- a/lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
+++ b/lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
@@ -110,6 +110,10 @@ public class Pauser {
       throw new PauserException("Failed to initialize the request coordinator.", e);
     }
 
+    // From here, we cannot throw exceptions right after they occur because we need to take care of
+    // the unpause operation failure. We will throw the exception after the unpause operation or at
+    // the end of this method.
+
     // Run a pause operation.
     PausedDuration pausedDuration = null;
     PauseFailedException pauseFailedException = null;
@@ -158,6 +162,12 @@ public class Pauser {
     // Create an error message if any of the operations failed.
     String errorMessage =
         createErrorMessage(unpauseFailedException, pauseFailedException, isTargetStatusEqual);
+
+    // We use the exceptions as conditions instead of using boolean flags like `isPauseOk`, etc. If
+    // we use boolean flags, it might cause a bit large number of combinations. For example, if we
+    // have three flags, they generate 2^3 = 8 combinations. It also might make the nested if
+    // statement or a lot of branches of the switch statement. That's why we don't use status flags
+    // for now.
 
     // Return the final result based on each process.
     if (unpauseFailedException != null) { // Unpause Failed.
--- a/lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
+++ b/lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
@@ -93,7 +93,7 @@ public class Pauser {
       throw new PauserException("Failed to initialize the request coordinator.", e);
     }
 
-    // Run pause operation.
+    // Run a pause operation.
     PauseFailedException pauseFailedException = null;
     try {
       pauseInternal(requestCoordinator, pauseDuration, maxPauseWaitTime);
@@ -101,7 +101,7 @@ public class Pauser {
       pauseFailedException = new PauseFailedException("Pause operation failed.", e);
     }
 
-    // Run unpause operation.
+    // Run an unpause operation.
     UnpauseFailedException unpauseFailedException = null;
     try {
       unpauseWithRetry(requestCoordinator, MAX_UNPAUSE_RETRY_COUNT);
@@ -168,7 +168,7 @@ public class Pauser {
     // during the pause duration.
     boolean backupOk = (pauseFailedException == null) && compareTargetSuccessful;
 
-    // Create error message if any of the operations failed.
+    // Create an error message if any of the operations failed.
     StringBuilder errorMessageBuilder = new StringBuilder();
     if (unpauseFailedException != null) {
       errorMessageBuilder.append(unpauseErrorMessage);
--- a/lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
+++ b/lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
@@ -36,8 +36,6 @@ public class Pauser {
   @VisibleForTesting static final int MAX_UNPAUSE_RETRY_COUNT = 3;
 
   private final TargetSelector targetSelector;
-  private Instant startTime;
-  private Instant endTime;
 
   /**
    * @param namespace The namespace where the pods are deployed.
@@ -94,9 +92,10 @@ public class Pauser {
     }
 
     // Run a pause operation.
+    PausedDuration pausedDuration = null;
     PauseFailedException pauseFailedException = null;
     try {
-      pauseInternal(requestCoordinator, pauseDuration, maxPauseWaitTime);
+      pausedDuration = pauseInternal(requestCoordinator, pauseDuration, maxPauseWaitTime);
     } catch (Exception e) {
       pauseFailedException = new PauseFailedException("Pause operation failed.", e);
     }
@@ -131,10 +130,8 @@ public class Pauser {
         "The target pods were updated during the pause duration. You cannot use the backup that"
             + " was taken during this pause duration. ";
     String unpauseFailedButPauseOkErrorMessage =
-        String.format(
-            "Note that the pause operations for taking backup succeeded. You can use a backup that"
-                + " was taken during this pause duration: Start Time = %s, End Time = %s. ",
-            startTime, endTime);
+        "Note that the pause operations for taking backup succeeded. You can use a backup that"
+            + " was taken during this pause duration. ";
 
     // Get pods and deployment information after pause.
     TargetSnapshot targetAfterPause;
@@ -196,7 +193,7 @@ public class Pauser {
       // issue might be caused by temporary issue, for example, pod restarts.
       throw new PauseFailedException(errorMessage);
     } else { // All operations succeeded.
-      return new PausedDuration(startTime, endTime);
+      return pausedDuration;
     }
   }
 
@@ -229,13 +226,13 @@ public class Pauser {
   }
 
   @VisibleForTesting
-  void pauseInternal(
+  PausedDuration pauseInternal(
       RequestCoordinator requestCoordinator, int pauseDuration, @Nullable Long maxPauseWaitTime) {
-
     requestCoordinator.pause(true, maxPauseWaitTime);
-    startTime = Instant.now();
+    Instant startTime = Instant.now();
     Uninterruptibles.sleepUninterruptibly(pauseDuration, TimeUnit.MILLISECONDS);
-    endTime = Instant.now();
+    Instant endTime = Instant.now();
+    return new PausedDuration(startTime, endTime);
   }
 
   @VisibleForTesting
--- a/lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
+++ b/lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
@@ -129,9 +129,6 @@ public class Pauser {
     String statusDifferentErrorMessage =
         "The target pods were updated during the pause duration. You cannot use the backup that"
             + " was taken during this pause duration. ";
-    String unpauseFailedButPauseOkErrorMessage =
-        "Note that the pause operations for taking backup succeeded. You can use a backup that"
-            + " was taken during this pause duration. ";
 
     // Get pods and deployment information after pause.
     TargetSnapshot targetAfterPause;
@@ -160,17 +157,10 @@ public class Pauser {
       }
     }
 
-    // If both the pause operation and status check succeeded, you can use the backup that was taken
-    // during the pause duration.
-    boolean isPauseOk = (pauseFailedException == null) && isTargetStatusEqual;
-
     // Create an error message if any of the operations failed.
     StringBuilder errorMessageBuilder = new StringBuilder();
     if (unpauseFailedException != null) {
       errorMessageBuilder.append(unpauseErrorMessage);
-      if (isPauseOk) {
-        errorMessageBuilder.append(unpauseFailedButPauseOkErrorMessage);
-      }
     }
     if (pauseFailedException != null) {
       errorMessageBuilder.append(pauseErrorMessage);
@@ -181,17 +171,18 @@ public class Pauser {
     String errorMessage = errorMessageBuilder.toString();
 
     // Return the final result based on each process.
-    if (unpauseFailedException
-        != null) { // Unpause issue is the most critical because it might cause system down.
+    if (unpauseFailedException != null) { // Unpause Failed.
+      // Unpause issue is the most critical because it might cause system down.
       throw new UnpauseFailedException(errorMessage, unpauseFailedException);
-    } else if (pauseFailedException
-        != null) { // Pause Failed is second priority because pause issue might be caused by
-      // configuration error.
+    } else if (pauseFailedException != null) { // Pause Failed.
+      // Pause Failed is second priority because pause issue might be caused by configuration error.
       throw new PauseFailedException(errorMessage, pauseFailedException);
-    } else if (!isTargetStatusEqual) { // Status check failed is third priority because this
-      // issue might be caused by temporary issue, for example, pod restarts.
+    } else if (!isTargetStatusEqual) { // Target status changed.
+      // Target status changed is third priority because this issue might be caused by temporary
+      // issue, for example, pod restarts.
       throw new PauseFailedException(errorMessage);
-    } else { // All operations succeeded.
+    } else {
+      // All operations succeeded.
       return pausedDuration;
     }
   }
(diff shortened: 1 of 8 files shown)


lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -84,17 +85,30 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+      endTime = Instant.now();
 
-    Instant endTime = Instant.now();
+      unpauseWithRetry(coordinator, MAX_UNPAUSE_RETRY_COUNT, target);
 
-    unpauseWithRetry(coordinator, MAX_UNPAUSE_RETRY_COUNT, target);
+    } catch (Exception e) {
+      unpauseWithRetry(coordinator, MAX_UNPAUSE_RETRY_COUNT, target);
+      throw e;
+    }
```
kota2and3kan, line 111: This is the main update of this PR.

Previously, there was no exception handling. So, there is a possibility that the Scalar products pod will keep `paused` status forever. In such a case, the Scalar product does not accept any request. It causes the service down.

To avoid such a case as long as possible, we want to run `unpause()` if any errors occur in any parts of the process.

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -79,26 +80,50 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
 
     TargetSnapshot target;
     try {
-      target = targetSelector.select();
+      target = getTarget();
```
kota2and3kan, line 83: I separated this `targetSelector.select()` as one method `getTarget()` for testing. This targetSelector is initialized in the constructor of `Pauser`. So, controlling the behavior of `targetSelector` by using Mockito is a bit complex and difficult. That's why I separate this function.

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -79,26 +80,50 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+      } catch (Exception ex) {
+        throw new PauserException(
+            "unpauseWithRetry() method failed twice due to unexpected exception.", e);
+      }
+      throw new PauserException(
+          "The pause operation failed for some reason. However, the unpause operation succeeded"
+              + " afterward. Currently, the scalar products are running with the unpause status."
+              + " You should retry the pause operation to ensure proper backup.",
+          e);
+    }
```
kota2and3kan, line 122: This is the main update of this PR.

Previously, there was no exception handling. So, if some error occurred, this library didn't try to run `unpause()`. This might cause the Scalar products to keep the pause state. In other words, it might cause like system down.

To address such issue, I added the exception handling here. In this implementation, if some errors occur, this library tries to run the `unpause()` operation.

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -123,17 +149,33 @@ private void unpauseWithRetry(
...
               target.getDeployment().getMetadata().getName());
-          return;
+          // In our CLI, we catch this exception and output the message as an error on the CLI side.
+          throw new PauserException(
+              String.format(
+                  "Failed to unpause Scalar product. They are still in paused. You must restart"
+                      + " related pods by using the `kubectl rollout restart deployment %s` command"
+                      + " to unpause all pods.",
+                  target.getDeployment().getMetadata().getName()),
+              e);
```
kota2and3kan, line 168: If `unpause()` failed three times, throw the `PauserException`. The caller of the `pause()` method should handle this exception.

I will update the message after I implement the `pauser.unpause()` method. In other words, I will add a new option (or subcommand) that allows users to run the unpause operation solely, in another PR.

After I implement such a new feature, users can use it to unpause Scalar products instead of using the `kubectl rollout restart deployment` command.

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -123,17 +149,33 @@ private void unpauseWithRetry(
...
         }
       }
     }
   }
 
+  @VisibleForTesting
+  TargetSnapshot getTarget() throws PauserException {
+    return targetSelector.select();
+  }
```
kota2and3kan, line 177: As I mentioned in another comment, I separated this `targetSelector.select()` as one method `getTarget()` for testing.

lib/src/test/java/com/scalar/admin/kubernetes/PauserTest.java
```diff
@@ -0,0 +1,478 @@
...
+
+class PauserTest {
+
+  private MockedStatic<Config> mockedConfig;
+  private RequestCoordinator requestCoordinator;
+  private TargetSnapshot targetSnapshot;
+  private TargetSelector targetSelector;
+  private V1Deployment deployment;
+  private V1ObjectMeta objectMeta;
+  private final String dummyObjectName = "dummyObjectName";
```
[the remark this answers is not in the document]
kota2and3kan: Thank you for your suggestion!
Fixed in f0870772cceb576f2df988d1f7acb870fb83d847.

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +77,163 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+    // Get pods and deployment information after pause.
     TargetSnapshot targetAfterPause;
     try {
-      targetAfterPause = targetSelector.select();
+      targetAfterPause = getTarget();
     } catch (Exception e) {
       throw new PauserException(
           "Failed to find the target pods to examine if the targets pods were updated during"
               + " paused.",
           e);
```
brfrn169, line 120: If pausing or unpausing fails, shouldn't we display a message about it here?

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +77,163 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+    try {
+      compareTargetSuccessful = compareTargetStatus(targetBeforePause, targetAfterPause);
+    } catch (Exception e) {
+      statusCheckFailedException = new StatusCheckFailedException(statusCheckErrorMessage, e);
+      if (unpauseFailedException == null) {
+        throw statusCheckFailedException;
+      } else {
+        throw new UnpauseFailedException(unpauseErrorMessage, statusCheckFailedException);
+      }
     }
```
brfrn169, line 159: It looks like `compareTargetStatus()` doesn't throw any exceptions. Do we still need this exception handling?

Maybe we need the exception handling for `targetAfterPause` here?
https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/33/files#diff-529e2e818d31d239079b031928623cab070acd5743e317b3d709c34e6e52b394R114-R121

brfrn169 (commented):
Left a few comments. Please take a look when you have time!

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
kota2and3kan: Ah, I see. You're right. We need to return the exception with messages here if the pause or unpause fails (especially, the unpause failure is critical).

I will fix it. Thank you for pointing it out!

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
kota2and3kan: > It looks like `compareTargetStatus()` doesn't throw any exceptions. Do we still need this exception handling?

For safety, I think it would be better to handle exceptions here. There is a possibility that some unexpected exceptions might occur here, and it could cause the issue `keeping unpause state`. 

> Maybe we need the exception handling for `targetAfterPause` here? https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/33/files#diff-529e2e818d31d239079b031928623cab070acd5743e317b3d709c34e6e52b394R114-R121

As I mentioned in another comment, you're right. I will add the exception handling there. I overlooked it. Thank you for pointing it out!

However, as mentioned above, I think we can keep this exception handling here. What do you think?

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
kota2and3kan: Fixed in d3b54c092b47410dfc9383424f9b7b608df1df7e.

kota2and3kan:
@brfrn169  
Thank you for your review!
I added the exception handling based on your feedback in d3b54c092b47410dfc9383424f9b7b608df1df7e, and I left comments.
Please take a look again when you have time! 🙇‍♂️

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +77,169 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
 
-    return new PausedDuration(startTime, endTime);
+    // If both the pause operation and status check succeeded, you can use the backup that was taken
+    // during the pause duration.
+    boolean backupOk = (pauseFailedException == null) && compareTargetSuccessful;
+
```
feeblefakie, line 170: I think it's more readable if we identify the cases here like the following and create error messages for each case.

```
boolean isAllOK = ...
boolean isPauseOkUnpauseNotOk = ...
boolean isPauseNotOkUnpauseOk = ...
boolean isPauseNotOkUnpauseNotOk = ...
```

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +77,169 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+    // Create error message if any of the operations failed.
+    StringBuilder errorMessageBuilder = new StringBuilder();
```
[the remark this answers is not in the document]
kota2and3kan: Ah, I see. I agree with your suggestion.
I will update it. Thank you!

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +77,169 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+        throw statusCheckFailedException;
+      } else {
+        throw new UnpauseFailedException(unpauseErrorMessage, statusCheckFailedException);
+      }
     }
 
-    return new PausedDuration(startTime, endTime);
+    // If both the pause operation and status check succeeded, you can use the backup that was taken
+    // during the pause duration.
+    boolean backupOk = (pauseFailedException == null) && compareTargetSuccessful;
```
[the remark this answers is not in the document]
kota2and3kan: I see. Indeed, it does not mean `Backup succeeded`. I will update the variable name.
Thank you for your suggestion!

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +77,169 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+  @VisibleForTesting
   RequestCoordinator getRequestCoordinator(TargetSnapshot target) {
     return new RequestCoordinator(
         target.getPods().stream()
             .map(p -> new InetSocketAddress(p.getStatus().getPodIP(), target.getAdminPort()))
             .collect(Collectors.toList()));
   }
+
+  @VisibleForTesting
+  void pauseInternal(
```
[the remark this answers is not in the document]
kota2and3kan: I see. It looks good!
I will update it. Thank you for your suggestion!

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +77,169 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+    // Run unpause operation.
+    UnpauseFailedException unpauseFailedException = null;
+    try {
+      unpauseWithRetry(requestCoordinator, MAX_UNPAUSE_RETRY_COUNT);
+    } catch (Exception e) {
+      unpauseFailedException = new UnpauseFailedException("Unpause operation failed.", e);
+    }
 
+    // Prepare error messages for each process.
+    String unpauseErrorMessage =
```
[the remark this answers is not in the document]
kota2and3kan: Ah, yes. That's right. They are fixed values.
I will update it. Thank you!

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -79,26 +80,50 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
-    Uninterruptibles.sleepUninterruptibly(pauseDuration, TimeUnit.MILLISECONDS);
+      Uninterruptibles.sleepUninterruptibly(pauseDuration, TimeUnit.MILLISECONDS);
 
-    Instant endTime = Instant.now();
+      endTime = Instant.now();
 
-    unpauseWithRetry(coordinator, MAX_UNPAUSE_RETRY_COUNT, target);
+      unpauseWithRetry(coordinator, MAX_UNPAUSE_RETRY_COUNT, target);
+
+    } catch (Exception e) {
```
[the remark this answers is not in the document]
komamitsu: Per another discussion, we concluded it was hard to simply the logic as I suggested due to a bit complex error handling.

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +77,169 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+      }
     }
 
-    if (!target.getStatus().equals(targetAfterPause.getStatus())) {
-      throw new PauserException("The target pods were updated during paused. Please retry.");
+    // Check if pods and deployment information are the same between before pause and after pause.
+    boolean compareTargetSuccessful;
+    StatusCheckFailedException statusCheckFailedException;
+    try {
+      compareTargetSuccessful = compareTargetStatus(targetBeforePause, targetAfterPause);
```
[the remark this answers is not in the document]
kota2and3kan: Oh, I see. I didn't know that.
I will update it. Thank you!

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
kota2and3kan: Ah, I see. Indeed, your suggestion is more readable.

However, in the previous meeting, we (@komamitsu  and I) discussed using boolean flags, and the result was `Don't use flags (boolean variables) as long as possible to reduce status/variables so that we can make code simpler`. So, I used the exceptions in the condition of the if statement.

Honestly, it's a bit difficult for me to decide which is better because I think both (more readable and simpler) make sense to me...

@komamitsu 
What do you think?

komamitsu: > `boolean isPauseOkUnpauseNotOk = ...`

This may improve the readability. But it'd also introduce many states. So, I think it's a trade-off (or maybe matter of taste?).

The current code uses `unpauseFailedException`, `pauseFailedException` and `compareTargetSuccessful` as more informative alternative to simple boolean flags. If we introduces new flags flags like `isPauseNotOkUnpauseOkCompareTargetNotOk`  (or state enum like `enum PauseResult { SUCCESS, PAUSE_SUCCESS_AND_UNPAUSE_FAILED_AND_COMPARE_TARGET_SUCCESS, ... }`?), we need to manage 8 (2 ** 3) states and it would make the method complex. So, I suggested to separately manage the 3 states as a simple way.

@kota2and3kan I think either is fine with us. I'm not insisting on my suggestion. You can go with any option you feel most confident in and comfortable being responsible for.

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +77,169 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+            + " was taken during this pause duration. ";
+    String unpauseFailedButBackupOkErrorMessage =
+        String.format(
+            "Note that the pause operations for taking backup succeeded. You can use a backup that"
+                + " was taken during this pause duration: Start Time = %s, End Time = %s. ",
+            startTime, endTime);
+
+    // Get pods and deployment information after pause.
     TargetSnapshot targetAfterPause;
+    PauserException pauserException;
```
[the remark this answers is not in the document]
kota2and3kan: Ah, yes. I can move it to the `catch` block.
I will update it. Thank you for pointing it out!

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +77,169 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+      } else {
+        throw new UnpauseFailedException(unpauseErrorMessage, pauserException);
+      }
     }
 
-    if (!target.getStatus().equals(targetAfterPause.getStatus())) {
-      throw new PauserException("The target pods were updated during paused. Please retry.");
+    // Check if pods and deployment information are the same between before pause and after pause.
+    boolean compareTargetSuccessful;
+    StatusCheckFailedException statusCheckFailedException;
```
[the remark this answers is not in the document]
kota2and3kan: Yes. We can move it to the `catch` block.
I will do it. Thank you!

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
kota2and3kan: I see. I got it.
I will consider which is better.
Thank you for your comment!

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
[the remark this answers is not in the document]
kota2and3kan: Fixed in c413386fb15e3ad65f7339694f0c134ea31dd347.

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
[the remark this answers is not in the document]
kota2and3kan: Fixed in c413386fb15e3ad65f7339694f0c134ea31dd347.

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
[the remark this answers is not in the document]
kota2and3kan: Fixed in 891beb85703faec5f8d82c18fb1892809d287a63.

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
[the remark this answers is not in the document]
kota2and3kan: Fixed in 744da27f63bfaad0aec89d157cd7146ec851c552.

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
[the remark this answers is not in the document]
kota2and3kan: Fixed in 9e9132dd6c0691164b1423f7edd7807889b12a49.

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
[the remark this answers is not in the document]
kota2and3kan: Fixed in 9e9132dd6c0691164b1423f7edd7807889b12a49.

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
kota2and3kan: Thank you for the discussion in the meeting!
Based on the result of the discussion, I updated as follows:

- Make the pause logic a bit simpler in 600e7b19fcf4592a0ca3eaa89af38d82fcd815fd to reduce many conditions and branches.
- Add code comments about our current decision in 909e596fcd0e1772ec5e91b55b34073b822f0ed8 for future development.

kota2and3kan:
@feeblefakie @komamitsu 
Thank you for your review and discussions!
I updated this PR based on feedback and discussion results.
Please take a look again when you have time! 🙇‍♂️

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +94,155 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
-      throw new PauserException(
-          "Failed to find the target pods to examine if the targets pods were updated during"
-              + " paused.",
-          e);
+      PauserException pauserException =
+          new PauserException(GET_TARGET_AFTER_PAUSE_ERROR_MESSAGE, e);
+      if (unpauseFailedException == null) {
+        throw pauserException;
+      } else {
+        throw new UnpauseFailedException(UNPAUSE_ERROR_MESSAGE, pauserException);
```
komamitsu, line 144: This code seems to drop the cause of `unpauseFailedException`. How about just adding `pauserException` to the `unpauseFailedException` as a suppressed one?

```java
    // Run an unpause operation.
    UnpauseFailedException unpauseFailedException = null;
    try {
      unpauseWithRetry(requestCoordinator, MAX_UNPAUSE_RETRY_COUNT);
    } catch (Exception e) {
      unpauseFailedException = new UnpauseFailedException(UNPAUSE_ERROR_MESSAGE, e);
    }

    // Get pods and deployment information after pause.
    TargetSnapshot targetAfterPause;
    try {
      targetAfterPause = getTarget();
    } catch (Exception e) {
      PauserException pauserException =
          new PauserException(GET_TARGET_AFTER_PAUSE_ERROR_MESSAGE, pauserException);
      if (unpauseFailedException == null) {
        throw pauserException;
      } else {
        unpauseFailedException.addSuppressed(e);
        throw unpauseFailedException;
      }
    }
```

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +94,155 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+      if (unpauseFailedException == null) {
+        throw statusCheckFailedException;
+      } else {
+        throw new UnpauseFailedException(UNPAUSE_ERROR_MESSAGE, statusCheckFailedException);
+      }
     }
 
-    return new PausedDuration(startTime, endTime);
+    // Create an error message if any of the operations failed.
+    String errorMessage =
```
komamitsu, line 163: [just an idea] Sorry for repeated minor suggestions. But, I just noticed adding additional exceptions to the primary exception might be an option so that all the detailed informations are preserved, although it would drop the benefit of outputting a fully concatenated error message in the first exception.

Like this:
```java
    ...

    Exception exception = buildException(pauseFailedException, unpauseFailedException,
        statusCheckException, statusUnmatchedException);
    if (exception != null) {
      throw exception;
    }
    return pausedDuration;
  }
  
  @Nullable
  private Exception buildException(
      @Nullable PauseFailedException pauseFailedException,
      @Nullable UnpauseFailedException unpauseFailedException,
      @Nullable StatusCheckFailedException statusCheckFailedException,
      @Nullable StatusUnmatchedException statusUnmatchedException) {
    Exception exception = null;
    if (unpauseFailedException != null) {
      exception = unpauseFailedException;
    }
    if (pauseFailedException != null) {
      if (exception == null) {
        exception = pauseFailedException;
      } else {
        exception.addSuppressed(pauseFailedException);
      }
    }
    if (statusCheckFailedException != null) {
      if (exception == null) {
        exception = statusCheckFailedException;
      } else {
        exception.addSuppressed(statusCheckFailedException);
      }
    }
    if (statusUnmatchedException != null) {
      if (exception == null) {
        exception = statusUnmatchedException;
      } else {
        exception.addSuppressed(statusUnmatchedException);
      }
    }
    return exception;
  }
```

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +94,155 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+    // statement or a lot of branches of the switch statement. That's why we don't use status flags
+    // for now.
+
+    // Return the final result based on each process.
+    if (unpauseFailedException != null) { // Unpause Failed.
+      // Unpause issue is the most critical because it might cause system down.
+      throw new UnpauseFailedException(errorMessage, unpauseFailedException);
+    } else if (pauseFailedException != null) { // Pause Failed.
+      // Pause Failed is second priority because pause issue might be caused by configuration error.
+      throw new PauseFailedException(errorMessage, pauseFailedException);
```
komamitsu, line 178: The first `PauseFailedException` instance contains a `PauseFailedException` instance as a cause. Is it expected? How about making `pauseFailedException` just Exception not `PauseFailedException`?

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
monkey-mas: +1 to the idea ;)

With this, API clients can decide what to do based on the primary + secondary(suppressed) exceptions, e.g., primary = `UnpauseFailedException`, secondary = `[StatusUnmatchedException]`.

@komamitsu Just wanted to confirm, you'd like to throw `PauserExcpetion` instead of `Exception`? 👀 Since all the  exception types are subclasses of `PauserExcpetion` (e.g., `PauseFailedException extends PauserExcpetion`), I _think_ it's ok to throw `PauserExcpetion` in this case? 🙄 (Or maybe I'm missing something :-)

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +94,155 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+    PauseFailedException pauseFailedException = null;
+    try {
+      pausedDuration = pauseInternal(requestCoordinator, pauseDuration, maxPauseWaitTime);
+    } catch (Exception e) {
+      pauseFailedException = new PauseFailedException("Pause operation failed.", e);
+    }
 
-    unpauseWithRetry(coordinator, MAX_UNPAUSE_RETRY_COUNT, target);
+    // Run an unpause operation.
+    UnpauseFailedException unpauseFailedException = null;
```
monkey-mas, line 127: @kota2and3kan 
I have a newbie question - just wanted to know why we execute `unpause()` even if `pause()` fails, e.g., when `pauseFailedException` is set? :eyes:

Is it probably because there might be some cluster nodes that are paused & running when `pause()` (partially) fails? Thus, one way or another, we have to run `unpause()` to restart all the nodes in cluster just in case? Do I have it correct?

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
komamitsu: @monkey-mas Good point. Agreed, although `StatusUnmatchedException` doesn't exist at the moment 😁

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
kota2and3kan: @monkey-mas 
Thank you for your review!

> Is it probably because there might be some cluster nodes that are paused & running when `pause()` (partially) fails? Thus, one way or another, we have to run `unpause()` to restart all the nodes in cluster just in case? Do I have it correct?

Yes. You're right.

If `pause()` failed, there is a possibility that some of the Pods of Scalar products are in a paused state.

In such a case, we should unpause those Pods by running `unpause()`.

The worst case is `pause()` pauses all Pods, but fails for some reason. It causes an issue like `system down` because Scalar product pods don't process client requests. Although it's a very rare case, but we need to run `unpause()` to unpause Scalar product pods to make they process client requests.

That's why we want to run `unpause()` after `pause()` fails. It's like a just-in-case operation.

FYI: Actually, `unpause()` doesn't restart pods. It just runs gRPC requests to unpause Scalar products. If `unpause()` fails as well as `pause()`, we should restart pods by using another way. It's a very rare case, but it occurs, for example, a network issue occurs.

monkey-mas: Cool, cool - thanks for your knowledge transfer :)

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
kota2and3kan: @komamitsu @monkey-mas 
Thank you for your review and suggestions!
It looks good to me!
I'm fixing it now. Please give me a bit of time.

(I didn't know `Throwable.addSuppressed()`. I was able to learn one new thing about Java!)

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
kota2and3kan: @komamitsu 
Ah, I see. Indeed, it drops the cause of `unpauseFailedException`... I overlooked it...

I think this issue should be resolved on the `buildException()` side, which we are discussing in another comment. So, I will fix it in the `buildException()` process.

Thank you for pointing it out!

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
kota2and3kan: @komamitsu 
Ah, yes. You're right. I think we can throw the `pauseFailedException ` as is instead of `new PauseFailedException()`.

I will fix it. I think we can fix this issue on the `buildException()` side that we are discussing in another comment.

Thank you for pointing it out!

lib/src/test/java/com/scalar/admin/kubernetes/PauserTest.java
```diff
@@ -0,0 +1,1599 @@
...
+              RuntimeException.class,
+              () -> pauser.unpauseWithRetry(requestCoordinator, MAX_UNPAUSE_RETRY_COUNT));
+      verify(requestCoordinator, times(MAX_UNPAUSE_RETRY_COUNT)).unpause();
+    }
+  }
+
+  @Nested
+  class buildException {
+    @Test
+    void buildException_00000_ReturnNull() throws PauserException {
```
kota2and3kan, line 610: To represent the combinations of arguments of `buildException()`, I used binary numbers like `00000`, `00001`, `00010`, `00011`, `00100`, `00101`... In this test, `1` means the exception is not null, and `0` means the exception is null. For example:

- `00101` means:
  - UnpauseFailedException is null
  - PauseFailedException is null
  - GetTargetAfterPauseFailedException is `NOT` null
  - StatusCheckFailedException is null
  - StatusUnmatchedException is `NOT` null

However, I don't know if this is good naming... So, I want to know some ideas about the test method naming, if you have some ideas. 😃

kota2and3kan:
@komamitsu @monkey-mas 
Thank you for your review!
I fixed code based on feedback.
Please take a look again when you have time! 🙇‍♂️

brfrn169 (approved):
LGTM! Thank you!

lib/src/test/java/com/scalar/admin/kubernetes/PauserTest.java
monkey-mas: > binary numbers `00000`

Good try :)
Although your strategy guarantees full coverage, the downside of it is the cost of maintainability :sweat:  Say, we'd add a new `PauserException`s - the total number of tests becomes 2^6 (64!!!), which is a bit over control for us :open_mouth: 

I'd say we could write some property-based tests for this so that we do't have to write that sort of full coverage tests. :-) i.e., instead of writing all the test cases, we can _generate_ potential (random) test cases & assert them.

## `buildException` spec 
IIUC, we are confident enough as long as we have the following tests scenarios in `PauserTest#buildException`:
- Returns `null` when all the exceptions are `null`.
- Returns `UnpauseFailedException` (+ suppressed exceptions) where `UnpauseFailedException` is not `null`.
- Returns `PauseFailedException` (+ suppressed exceptions) where
  - a) `PauseFailedException` is not `null`.
  - b) `UnpauseFailedException` is `null`.
- Returns `GetTargetAfterPauseFailedException` (+ suppressed exceptions) where
  - a) `GetTargetAfterPauseFailedException` is not `null`
  - b) both `UnpauseFailedException` and `PauseFailedException`  are `null`.
- Returns `StatusCheckFailedException` (+ suppressed exceptions) where
  - a) `StatusCheckFailedException` is not `null`
  - b) `UnpauseFailedException`, `PauseFailedException` and `GetTargetAfterPauseFailedException` are `null`.
- Returns `StatusUnmatchedException` (+ suppressed exceptions) where
  - a) `StatusUnmatchedException` is not `null`
  - b) all the other `PauserException` are `null`.

## PB test strategy
The spec indicates that we just need to write `N+1` tests where `N` is the number of exceptions, which could be better than `2^N` full-coverage tests due to maintainability :) This means that we don't have to write or maintain a lot of test cases even when we add a new `PauserException` in the future!

### How `setup` (generate) test data?
1. Randomly generate `PauserException`s for all the exceptions for `buildException(...)`.
  - The main purpose of this is to generate `suppressed` (secondary) exceptions.
  - This could be done in `@BeaforeEach`
2. Generate and set a primary `PauserException` in each test case.
  - Make sure that nullify unnecessary exception.
  - For example, we must assign `null` to `UnpauseFailedException` to obtain `PauseFailedException` as a primary exception.
3. Assert that an expected primary & secondary exceptions are obtained.

monkey-mas: # JUnit test code (just an idea)
@kota2and3kan @komamitsu 
The following is my thoughts... Your feedback is appreciated :) 

## Config in Gradle
@kota2and3kan 
Actually... I used AssertJ to execute `assertThat(actual).contains(exccted)` and so on since it's handy for testing :sweat_smile: .

Please set up your Gradle as follows:
- `build.gradle` in `lib` module
```gradle
testImplementation("org.assertj:assertj-core:${assertjVersion}")
```
- `build.gradle` in top-level dir
```
assertjVersion = '3.25.3'
```
- Import AssertJ in `PauserTest`
```java
import static org.assertj.core.api.Assertions.assertThat;
```

## JUnit
[Memo] I used `@Repeated(X)` to do property-based testing for randomly-generated `suppressed` exceptions. `X` is the number of PB random tests :)

```java
  @Nested
  class buildException {
    private Pauser pauser;

    private UnpauseFailedException unpauseFailedException;
    private PauseFailedException pauseFailedException;
    private GetTargetAfterPauseFailedException getTargetAfterPauseFailedException;
    private StatusCheckFailedException statusCheckFailedException;
    private StatusUnmatchedException statusUnmatchedException;

    @BeforeEach
    void setUp() throws PauserException {
      String namespace = "dummyNs";
      String helmReleaseName = "dummyRelease";
      pauser = new Pauser(namespace, helmReleaseName);

      // Randomly generate exceptions.
      String dummyMessage = "dummyMessage";
      unpauseFailedException = random.nextBoolean() ? new UnpauseFailedException(dummyMessage) : null;
      pauseFailedException = random.nextBoolean() ? new PauseFailedException(dummyMessage) : null;
      getTargetAfterPauseFailedException = random.nextBoolean() ? new GetTargetAfterPauseFailedException(dummyMessage) : null;
      statusCheckFailedException = random.nextBoolean() ? new StatusCheckFailedException(dummyMessage) : null;
      statusUnmatchedException = random.nextBoolean() ? new StatusUnmatchedException(dummyMessage) : null;
    }

    @Nested
    class whereAllExceptionsAreNull {
      @Test
      void returnsNull() {
        // Arrange
        unpauseFailedException = null;
        pauseFailedException = null;
        getTargetAfterPauseFailedException = null;
        statusCheckFailedException = null;
        statusUnmatchedException = null;

        // Act
        Exception actual =
                pauser.buildException(
                        unpauseFailedException,
                        pauseFailedException,
                        getTargetAfterPauseFailedException,
                        statusCheckFailedException,
                        statusUnmatchedException);

        // Assert
        assertThat(actual).isNull();
      }
    }

    @Nested
    class whereUnpauseFailedExceptionIsSet {
      List<PauserException> expectedSecondaryExceptions;

      @BeforeEach
      void setUp() {
        unpauseFailedException = new UnpauseFailedException("Primary exception");

        PauserException[] pauserExceptions = new PauserException[]{
                pauseFailedException,
                getTargetAfterPauseFailedException,
                statusCheckFailedException,
                statusUnmatchedException
        };
        expectedSecondaryExceptions = Arrays.stream(pauserExceptions).filter(Objects::nonNull).collect(Collectors.toList());
      }

      @RepeatedTest(20)
      void returnsUnpausedFailedException() {
        // Act
        Exception actual =
                pauser.buildException(
                        unpauseFailedException,
                        pauseFailedException,
                        getTargetAfterPauseFailedException,
                        statusCheckFailedException,
                        statusUnmatchedException);

        // Assert
        assertThat(actual).isInstanceOf(UnpauseFailedException.class);
        assertThat(actual.getSuppressed()).containsAll(expectedSecondaryExceptions);
      }
    }

    @Nested
    class wherePauseFailedExceptionIsSet {
      @Nested
      class whenUnpauseFailedExceptionIsNull {
        List<PauserException> expectedSecondaryExceptions;

        @BeforeEach
        void setUp() {
          pauseFailedException = new PauseFailedException("Primary exception");

          // Arrange a situation where UnpauseFailedException is null.
          unpauseFailedException = null;

          PauserException[] pauserExceptions = new PauserException[]{
                  getTargetAfterPauseFailedException,
                  statusCheckFailedException,
                  statusUnmatchedException
          };
          expectedSecondaryExceptions = Arrays.stream(pauserExceptions).filter(Objects::nonNull).collect(Collectors.toList());
        }

        @RepeatedTest(10)
        void returnsPausedFailedException() {
          // Act
          Exception actual =
                  pauser.buildException(
                          unpauseFailedException,
                          pauseFailedException,
                          getTargetAfterPauseFailedException,
                          statusCheckFailedException,
                          statusUnmatchedException);

          // Assert
          assertThat(actual).isInstanceOf(PauseFailedException.class);
          assertThat(actual.getSuppressed()).containsAll(expectedSecondaryExceptions);
        }
      }
    }

    @Nested
    class whereGetTargetAfterPauseFailedExceptionIsSet {
      @Nested
      @DisplayName("where UnpauseFailedException and PauseFailedException are null")
      class whenUnpauseFailedExceptionAndPauseFailedExceptionAreNull {
        List<PauserException> expectedSecondaryExceptions;

        @BeforeEach
        void setUp() {
          getTargetAfterPauseFailedException = new GetTargetAfterPauseFailedException("Primary exception");

          // Arrange a situation where UnpauseFailedException and PauseFailedException are null.
          unpauseFailedException = null;
          pauseFailedException = null;

          PauserException[] pauserExceptions = new PauserException[]{
                  statusCheckFailedException,
                  statusUnmatchedException
          };
          expectedSecondaryExceptions = Arrays.stream(pauserExceptions).filter(Objects::nonNull).collect(Collectors.toList());
        }

        @RepeatedTest(5)
        void returnsGetTargetAfterPauseFailedException() {
          // Act
          Exception actual =
                  pauser.buildException(
                          unpauseFailedException,
                          pauseFailedException,
                          getTargetAfterPauseFailedException,
                          statusCheckFailedException,
                          statusUnmatchedException);

          // Assert
          assertThat(actual).isInstanceOf(GetTargetAfterPauseFailedException.class);
          assertThat(actual.getSuppressed()).containsAll(expectedSecondaryExceptions);
        }
      }
    }

    @Nested
    class whereStatusCheckFailedExceptionIsSet {
      @Nested
      @DisplayName("where UnpauseFailedException, PauseFailedException and GetTargetAfterPauseFailedException are null")
      class whenUnpauseFailedExceptionAndPauseFailedExceptionAndGetTargetAfterPauseFailedExceptionAreNull {
        List<PauserException> expectedSecondaryExceptions;

        @BeforeEach
        void setUp() {
          statusCheckFailedException = new StatusCheckFailedException("Primary exception");

          // Arrange a situation where UnpauseFailedException, PauseFailedException and GetTargetAfterPauseFailedException are null.
          unpauseFailedException = null;
          pauseFailedException = null;
          getTargetAfterPauseFailedException = null;

          PauserException[] pauserExceptions = new PauserException[]{statusUnmatchedException};
          expectedSecondaryExceptions = Arrays.stream(pauserExceptions).filter(Objects::nonNull).collect(Collectors.toList());
        }

        @RepeatedTest(5)
        void returnsStatusCheckFailedException() {
          // Act
          Exception actual =
                  pauser.buildException(
                          unpauseFailedException,
                          pauseFailedException,
                          getTargetAfterPauseFailedException,
                          statusCheckFailedException,
                          statusUnmatchedException);

          // Assert
          assertThat(actual).isInstanceOf(StatusCheckFailedException.class);
          assertThat(actual.getSuppressed()).containsAll(expectedSecondaryExceptions);
        }
      }
    }

    @Nested
    class whereStatusUnmatchedExceptionIsSet {
      @Nested
      @DisplayName("where all the other exceptions are null")
      class whenAllOtherExceptionsAreNull {
        @BeforeEach
        void setUp() {
          statusUnmatchedException = new StatusUnmatchedException("Primary exception");

          // Arrange a situation where all the other exceptions are null.
          unpauseFailedException = null;
          pauseFailedException = null;
          getTargetAfterPauseFailedException = null;
          statusCheckFailedException = null;
        }

        @Test
        void returns() {
          // Act
          Exception actual =
                  pauser.buildException(
                          unpauseFailedException,
                          pauseFailedException,
                          getTargetAfterPauseFailedException,
                          statusCheckFailedException,
                          statusUnmatchedException);

          // Assert
          assertThat(actual).isInstanceOf(StatusUnmatchedException.class);
          assertThat(actual.getSuppressed()).isEmpty();
        }
      }
    }
  }
```

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +103,188 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+  }
+
+  @VisibleForTesting
+  @Nullable
+  StatusUnmatchedException targetStatusEquals(
+      TargetSnapshot before, @Nullable TargetSnapshot after) {
+    if (after != null && before.getStatus().equals(after.getStatus())) {
+      return null;
+    } else {
+      return new StatusUnmatchedException(STATUS_UNMATCHED_ERROR_MESSAGE);
```
komamitsu, line 230: [minor] How about adding `before` and `after` to the error message just in case?

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +103,188 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+      getTargetAfterPauseFailedException =
+          new GetTargetAfterPauseFailedException(GET_TARGET_AFTER_PAUSE_ERROR_MESSAGE, e);
     }
 
-    if (!target.getStatus().equals(targetAfterPause.getStatus())) {
-      throw new PauserException("The target pods were updated during paused. Please retry.");
+    // Check if pods and deployment information are the same between before pause and after pause.
+    StatusCheckFailedException statusCheckFailedException = null;
+    StatusUnmatchedException statusUnmatchedException = null;
+    try {
```
komamitsu, line 156: How about executing this block only when `targetAfterPause != null`? The method seems to return `StatusUnmatchedException` when the argument `after` is null, but in this case comparing `before` and `after` itself isn't conducted.

kota2and3kan: Ah, yes. That's right. I will fix it. Thank you for the suggestion!

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
kota2and3kan: Ah, I see. It might have a lot of information that makes the error message very long, but I will check what is included in `before` and `after`, first. And, if the information is not too large, I will add them into the error message.

Thank you for your suggestion!

lib/src/test/java/com/scalar/admin/kubernetes/PauserTest.java
kota2and3kan: @monkey-mas 
Thank you for your review! This property-based test information is very helpful to me!

Totally, it looks good to me, but I want to progress this PR with the current unit test as is. This is because we prioritize the release of the Omnistrate service.

Therefore, I will fix the unit test in another PR in the future.

Anyway, thank you for your suggestion!

komamitsu: @monkey-mas Basically looks good to me! 👍 

I think the number of repeats can be larger like up to 50 since it won't take so much time because of no wait.

BTW, I'm not sure whether the nested class name should start with a lowercase...

komamitsu: > up to 50

Just noticed it's close to the full combination(64) 😁
Maybe full coverage test is also okay using `@ParameterizedTest`.

monkey-mas: Thanks for the feedback, folks :)



> Therefore, I will fix the unit test in another PR in the future.

Yep, understood 👌 

> I think the number of repeats can be larger like up to 50 since it won't take so much time because of no wait.

Yeah, true. I'm open to use a large number as you suggested :)

For some tests, the number does not have to be that big and I just set a smaller number 😓 For instance, in `StatusCheckFailedException` test, `statusUnmatchedException` is the only randomly generated variable. (Thus, _statistically speaking_, `5` was enough 😓)
ref. [Expected time to roll all 1 through 6 on a die](https://math.stackexchange.com/a/28906)

Anyway, you're right in the sense that it doesn't hurt even if we use `50` for that sort of unit test either :)

> BTW, I'm not sure whether the nested class name should start with a lowercase...

Yeah, I sort of got your point. 😅 

I actually don't follow Java naming convention here in unit tests 🙄 (eg a class name starts with a capital letter.)
- I usually create the following nested test class structure
  - The nested class name starts with a lower letter to match a method name (which almost always starts with a lower case letter in Java).
  - e.g., `buildException` but not `BuildException` to clarify the target method to test.
- and... I also apply the same rule for nested classes... 🤫 (except the top-level unit test name)

```java
class UnitTest {
  @Nested
  class methodNameToTest {
    // More nested classes or tests.
  }
}
```

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +103,188 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+            unpauseFailedException,
+            pauseFailedException,
+            getTargetAfterPauseFailedException,
+            statusCheckFailedException,
+            statusUnmatchedException);
+
+    // Return the final result based on each process.
+    if (pauserException != null) {
+      // Some operations failed.
+      throw pauserException;
```
monkey-mas, line 178: [minor] Should we have some log messages when `pauseException != null`? 👀 

Looks like we removed a log in this PR (Sorry if I miss some context here 😓)
```java
          logger.warn(
              "Failed to unpause Scalar product. They are still in paused. You must restart related"
                  + " pods by using the `kubectl rollout restart deployment {}`"
                  + " command to unpase all pods.",
              target.getDeployment().getMetadata().getName());
          return;
```

monkey-mas (approved):
LGTM 👍  Let's create follow-up PRs if necessary once this base PR is merged :)

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
kota2and3kan: I think we don't need to output such a log message here.

I decided to include all messages in the generated exception.

Now, we can show users clearer errors by using several kinds of exceptions (e.g., `UnpauseFailedException`).

Also, I wanted to see `What happened` from the log, but now we can see all the messages in the exceptions via the stacktrace.

So, there is no problem if we don't have a log here.

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
kota2and3kan: Fixed in 5cca44c080bffc253678d0dcb89ba3c9ec432d11.

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
kota2and3kan: I checked what is included in `TargetStatus` (i.e., `before` and `after`). As a result, I think we can add their information to the log, but in the current implementation, I think the information might not be readable a little.

So, I think it would be better to implement `TargetStatus.toString()` first. However, I think we need to consider something before implementing it.

Therefore, I will keep the current implementation as is in this PR, and I will fix it in another PR in the future.

lib/src/test/java/com/scalar/admin/kubernetes/PauserTest.java
kota2and3kan: I see. The naming is very difficult... 
Anyway, thank you for your comments! I will refer to it when I create another PR that fixes unit tests.

kota2and3kan:
@komamitsu 
Thank you for your review!
I applied your suggestion and left the answer to your comment.
Please take a look again when you have time! 🙇‍♂️

komamitsu (approved):
LGTM! 👍

feeblefakie (approved):
LGTM! Thank you!

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +103,189 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+      @Nullable StatusCheckFailedException statusCheckFailedException,
+      @Nullable StatusUnmatchedException statusUnmatchedException) {
+    PauserException pauserException = null;
+
+    // Unpause issue is the most critical because it might cause system down.
+    if (unpauseFailedException != null) {
+      pauserException = unpauseFailedException;
+    }
+
+    // Pause Failed is second priority because pause issue might be caused by configuration error.
```
feeblefakie, line 250:
```suggestion
    // Treat the pause failure as the second priority because the issue might be caused by a configuration mistake.
```

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +103,189 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+  @VisibleForTesting
+  PauserException buildException(
+      @Nullable UnpauseFailedException unpauseFailedException,
+      @Nullable PauseFailedException pauseFailedException,
+      @Nullable GetTargetAfterPauseFailedException getTargetAfterPauseFailedException,
+      @Nullable StatusCheckFailedException statusCheckFailedException,
+      @Nullable StatusUnmatchedException statusUnmatchedException) {
+    PauserException pauserException = null;
+
+    // Unpause issue is the most critical because it might cause system down.
```
feeblefakie, line 245:
```suggestion
    // Treat the unpause failure as most critical because it might cause system unavailability.
```

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +103,189 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+    // Pause Failed is second priority because pause issue might be caused by configuration error.
+    if (pauseFailedException != null) {
+      if (pauserException == null) {
+        pauserException = pauseFailedException;
+      } else {
+        pauserException.addSuppressed(pauseFailedException);
+      }
+    }
+
+    // Get target issue is third priority because this issue might be caused by temporary issue, for
```
feeblefakie, line 259:
```suggestion
    // Treat the getting target failure as the third priority because this issue might be caused by a temporary glitch, for
```

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +103,189 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+    // example, network issues.
```
feeblefakie, line 260:
```suggestion
    // example, network failures.
```

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +103,189 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+    if (getTargetAfterPauseFailedException != null) {
+      if (pauserException == null) {
+        pauserException = getTargetAfterPauseFailedException;
+      } else {
+        pauserException.addSuppressed(getTargetAfterPauseFailedException);
+      }
+    }
+
+    // Status check issue is third priority because this issue might be caused by temporary issue,
```
feeblefakie, line 269:
```suggestion
    // Treat the status checking failure as third priority because this issue might be caused by a temporary glitch,
```

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +103,189 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+    // for example, targetAfterPause is null.
```
feeblefakie, line 270: What causes `targetAfterPause is null`?

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +103,189 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+    if (statusCheckFailedException != null) {
+      if (pauserException == null) {
+        pauserException = statusCheckFailedException;
+      } else {
+        pauserException.addSuppressed(statusCheckFailedException);
+      }
+    }
+
+    // Status unmatched issue is third priority because this issue might be caused by temporary
```
feeblefakie, line 279:
```suggestion
    // Treat the status unmatched issue as third priority because this issue might be caused by temporary
```

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
```diff
@@ -77,67 +103,189 @@ public PausedDuration pause(int pauseDuration, @Nullable Long maxPauseWaitTime)
...
+    // issue, for example, pod restarts.
```
feeblefakie, line 280:
```suggestion
    // glitch, for example, pod restarts.
```

feeblefakie (commented):
Sorry, I left some minor suggestions on the English comments.

feeblefakie (changes requested):
Sorry, I left some suggestions.

lib/src/main/java/com/scalar/admin/kubernetes/Pauser.java
kota2and3kan: Ah, sorry. This comment does not have meaningful information.

For example, if we cannot get the pod or deployment information by using the Kubernetes API, the `targetAfterPause` will be `null`, and enter this conditional branch.

I will fix this comment. Thank you for your question!

kota2and3kan: Fixed in 639e132e4766b0299b8b91047b8b633f0371072f and b93c054f00233d3d37ebb1571c9031f6d68e7747.

kota2and3kan:
@feeblefakie 
Thank you for your review!
I fixed the code comments and left an answer to your question.
Please take a look again when you have time!

feeblefakie (approved):
LGTM! Thank you!

PR #46, opened 2026-06-01
Refactor to Onion Architecture with DDD Patterns

monkey-mas:
## Description

This PR refactors the scalar-admin-for-kubernetes codebase from a monolithic class-based approach to a proper layered architecture following **Domain-Driven Design (DDD)** and **Onion Architecture** principles.

**Why this refactoring is necessary:**

The original codebase had several architectural issues:
- **Monolithic design**: Business logic, infrastructure concerns, and presentation logic were mixed in `Pauser` and `TlsPauser` classes
- **Domain leakage**: Domain objects like `PauseDuration` were exposed directly to external layers (CLI)
- **Poor separation of concerns**: No clear boundaries between layers
- **Limited extensibility**: Difficult to add new pod discovery modes or pause strategies

This refactoring establishes clear layer boundaries, improves testability, and sets the foundation for future extensibility.
## Related issues and/or PRs

> If this PR addresses or references any issues and/or other PRs, list them here.

## Related issues and/or PRs

- Next step: #47 

### Before (Monolithic)
```
cli/Cli.java
  └─> lib/Pauser.java (monolithic class with everything)
       ├─> Business logic + Infrastructure + Error handling
       └─> TlsPauser.java (duplicate logic for TLS)
```

### After (Layered Architecture)
```
cli/ (module)
  └─> Cli.java (CLI interface)
       └─> presentation/PauseController.java (Presentation layer)

lib/ (module)
  ├─> presentation/
  │    └─> PauseController.java (Presentation layer)
  │         └─> application/
  │              ├─> PauseApplicationService.java (Application layer)
  │              └─> dto/
  │                   ├─> PauseDurationDto
  │                   └─> PauseRequest
  │
  ├─> domain/
  │    ├─> service/
  │    │    └─> PauseService.java (Domain service)
  │    ├─> model/
  │    │    ├─> pause/
  │    │    │    ├─> PauseCommand (sealed interface)
  │    │    │    ├─> PauseByHelmReleaseCommand (record)
  │    │    │    ├─> PauseTarget (aggregate root)
  │    │    │    ├─> PauseDuration (value object)
  │    │    │    └─> TlsConfig (value object)
  │    │    └─> shared/
  │    │         └─> Product (enum)
  │    ├─> client/
  │    │    ├─> KubernetesClient (interface)
  │    │    ├─> ScalarAdminClient (interface)
  │    │    └─> ScalarAdminClientFactory (interface)
  │    └─> exception/
  │         └─> PauserException (and subclasses)
  │
  └─> infrastructure/
       └─> client/
            ├─> KubernetesClientImpl
            ├─> ScalarAdminClientImpl
            ├─> TlsScalarAdminClientImpl
            └─> ScalarAdminClientFactoryImpl
```

Dependency Flow:
cli → lib/presentation → lib/application → lib/domain ← lib/infrastructure

#### 1. Domain Layer Foundation

Migrated core domain concepts to proper domain layer:
- Exception classes with comprehensive Javadoc
- Product enum with Java 21 switch expressions
- PauseDuration as immutable record with validation

#### 2. Aggregate Pattern

Unified `TargetSnapshot` and `TargetStatus` into `PauseTarget` aggregate root using Java 21 records, establishing clear boundaries and single source of truth for pause target state.

#### 3. Client Pattern

Refactored `TargetSelector` to Client pattern with domain interface (`KubernetesClient`) and infrastructure implementation (`KubernetesClientImpl`). Renamed `findByHelmRelease` → `resolvePauseTargetByHelmRelease` to better describe the complex multi-step resolution process.

#### 4. Command Pattern & Layered Services

Broke down monolithic `Pauser`/`TlsPauser` classes into proper layers:
- **Command**: Sealed interface `PauseCommand` with `PauseByHelmReleaseCommand` record
- **Domain Service**: `PauseService` for business logic and retry/error handling
- **Application Service**: `PauseApplicationService` for orchestration with switch expressions
- **Presentation**: `PauseController` for dependency initialization

Extracted `TlsConfig` as value object and introduced `ScalarAdminClient` interface with factory pattern. Applied DIP to `ScalarAdminClientFactory` by extracting interface to domain layer (`ScalarAdminClientFactory`) with infrastructure implementation (`ScalarAdminClientFactoryImpl`), consistent with `KubernetesClient`/`KubernetesClientImpl` pattern.  

#### 5. DTO Pattern

Introduced DTOs to prevent domain leakage:
- `PauseDurationDto`: Primitives-only record for cross-layer communication
- `PauseRequest`: Unified input validation and simplified Controller interface

Application layer now acts as anti-corruption layer, converting domain objects to DTOs at boundaries.

#### 6. Testing Improvements

Modernized test suite with AssertJ, `@Nested`, and `@DisplayName`. Improved test realism:
- Changed from 1 pod to 3 pods (realistic ScalarDB Cluster)
- Pod resource version correctly changes with restart count
- Realistic pod deletion/recreation scenarios

## Checklist

> The following is a best-effort checklist. If any items in this checklist are not applicable to this PR or are dependent on other, unmerged PRs, please still mark the checkboxes after you have read and understood each item.

- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have updated the documentation to reflect the changes.
- [ ] Any remaining open issues linked to this PR are documented and up-to-date (Jira, GitHub, etc.).
- [ ] Tests (unit, integration, etc.) have been added for the changes.
- [ ] My changes generate no new warnings.
- [ ] Any dependent changes in other PRs have been merged and published.

### Migration Path

This refactoring was done incrementally:

1. **Phase 1**: Migrate basic domain models (exceptions, Product, PauseDuration)
2. **Phase 2**: Introduce aggregate (PauseTarget)
3. **Phase 3**: Introduce client pattern (KubernetesClient, refactored from initial repository pattern)
4. **Phase 4**: Refactor services into layers (PauseService, PauseApplicationService, PauseController)
5. **Phase 5**: Introduce DTO pattern (PauseDurationDto, PauseRequest)
6. **Phase 6**: Refine naming (PauseTargetRepository → KubernetesClient)
7. **Phase 7**: Improve test realism and coverage

### Manual Integration Testing

To verify this refactoring works correctly, follow the manual steps below with minikube:

**Prerequisites:**

- minikube is running
- ScalarDB Cluster is already running in minikube (deployed via Helm as `scalardb-cluster` in `default` namespace)
- Helm repo `scalar-labs` is added (`helm repo add scalar-labs https://scalar-labs.github.io/helm-charts`)

**Important:** For Test 2b (container restart test) to work properly, ScalarDB Cluster must be configured with a short decommissioning duration. This PR includes the following change to `.github/manifests/scalardb-cluster.yaml`:

```yaml
scalardbCluster:
  scalardbClusterNodeProperties: |
    scalar.db.cluster.node.decommissioning_duration_secs=1
```

This setting allows `kill -15` to restart the container quickly within the pause window, enabling proper detection of restart count changes.

**Build and load the CLI image:** 

```sh
# Build CLI shadow JAR (fat JAR with all dependencies)
./gradlew clean :cli:shadowJar

# Build Docker image
cd cli && docker buildx build -t scalar-admin-for-kubernetes-cli:local-test -f Dockerfile . --load && cd ..

# Remove old image from minikube (if exists)
minikube image rm scalar-admin-for-kubernetes-cli:local-test || true

# Load new image into minikube
minikube image load scalar-admin-for-kubernetes-cli:local-test
```

**Test 1: Normal pause/unpause operation (success case)**

```sh
# Create Helm values file
cat > /tmp/test-values.yaml <<EOF
scalarAdminForKubernetes:
  image:
    repository: scalar-admin-for-kubernetes-cli
    tag: local-test
    pullPolicy: Never
  jobType: "job"
  commandArgs:
    - --release-name
    - scalardb-cluster
    - --namespace
    - default
    - --pause-duration
    - "3000"
EOF

# Deploy scalar-admin-for-kubernetes
helm install scalar-admin-test scalar-labs/scalar-admin-for-kubernetes \
    -f /tmp/test-values.yaml -n default

# Wait for job to complete
kubectl wait --for=condition=complete job -l app.kubernetes.io/app=scalar-admin-for-kubernetes \
    -n default --timeout=60s

# Check job logs (should show JSON output with pause times)
kubectl logs -l app.kubernetes.io/app=scalar-admin-for-kubernetes -n default | jq .

# Expected output:
# {
#   "namespace": "default",
#   "timezone": "Etc/UTC",
#   "helm_release_name": "scalardb-cluster",
#   "pause_start_timestamp_ms": 1780026465487,
#   "pause_end_timestamp_ms": 1780026468489,
#   "pause_start_date_time": "2026-05-29T03:47:45.487",                                                                                              
#   "pause_end_date_time": "2026-05-29T03:47:48.489"
# }

# Verify ScalarDB Cluster logs show pause/unpause messages
kubectl logs -l app.kubernetes.io/app=scalardb-cluster -n default --tail=50 | grep -i pause

# Expected output:
# 2026-05-29 03:47:45,480 [INFO  com.scalar.db.cluster.node.ScalarAdminGrpcService] Paused
# 2026-05-29 03:47:48,505 [INFO  com.scalar.db.cluster.node.ScalarAdminGrpcService] Unpaused

# Cleanup
helm uninstall scalar-admin-test -n default
```

**Test 2a: Pod deletion during pause (error case - pod name changed)**

When a pod is deleted during pause, the CLI detects the pod name change. However, since `ScalarAdminClient` caches pod IPs at pause start, the unpause operation also fails because the deleted pod's IP becomes unreachable. As a result, `UnpauseFailedException` is the primary error and `StatusUnmatchedException` appears as a suppressed exception. 

```sh
# Deploy with longer pause duration
cat > /tmp/test2a-values.yaml <<EOF
scalarAdminForKubernetes:
  image:
    repository: scalar-admin-for-kubernetes-cli
    tag: local-test
    pullPolicy: Never
  jobType: "job"
  commandArgs:
    - --release-name
    - scalardb-cluster
    - --namespace
    - default
    - --pause-duration
    - "60000"  # 60 seconds
EOF

helm install scalar-admin-test scalar-labs/scalar-admin-for-kubernetes \
    -f /tmp/test2a-values.yaml -n default

# Confirm pause has started by checking ScalarDB logs                                                                                                
kubectl logs -l app.kubernetes.io/app=scalardb-cluster -n default --since=120s | grep -i "Paused"

# Graceful delete one pod (Kubernetes will create a replacement with a different name)
POD_NAME=$(kubectl get pods -l app.kubernetes.io/app=scalardb-cluster -n default -o jsonpath='{.items[0].metadata.name}')
kubectl delete pod $POD_NAME -n default

# Wait for replacement pod to be Ready
kubectl wait --for=condition=Available deployment/scalardb-cluster-node -n default --timeout=60s

# Wait for job to fail
kubectl wait --for=condition=failed job -l app.kubernetes.io/app=scalar-admin-for-kubernetes,app.kubernetes.io/instance=scalar-admin-test \
    -n default --timeout=90s

# Check job logs
JOB_POD=$(kubectl get pod -l app.kubernetes.io/app=scalar-admin-for-kubernetes,app.kubernetes.io/instance=scalar-admin-test -n default -o jsonpath='{.items[0].metadata.name}')
kubectl logs $JOB_POD -n default 2>&1 | grep -E "(UnpauseFailedException|StatusUnmatchedException)"

# Expected output (primary: UnpauseFailedException, suppressed: StatusUnmatchedException):
# com.scalar.admin.kubernetes.domain.exception.UnpauseFailedException: Unpause operation failed. ...
#   Suppressed: com.scalar.admin.kubernetes.domain.exception.StatusUnmatchedException: The target pods were updated during the pause duration. ...

# Cleanup
helm uninstall scalar-admin-test -n default
# Restart deployment to unpause any remaining paused pods
kubectl rollout restart deployment scalardb-cluster-node -n default
```

**Test 2b: Container restart during pause (error case - restart count changed)**

NOTE: This test requires ScalarDB Cluster to be configured with `decommissioning_duration_secs=1` so that `kill -15` restarts the container quickly within the pause window.

When a container is restarted via `kill -15`, the pod name and IP remain unchanged (only the restart count increments). Therefore, the unpause operation succeeds and `StatusUnmatchedException` is the primary error.

```sh
# Deploy with longer pause duration
cat > /tmp/test2b-values.yaml <<EOF
scalarAdminForKubernetes:
  image:
    repository: scalar-admin-for-kubernetes-cli
    tag: local-test
    pullPolicy: Never
  jobType: "job"
  commandArgs:
    - --release-name
    - scalardb-cluster
    - --namespace
    - default
    - --pause-duration
    - "60000"  # 60 seconds
EOF

helm install scalar-admin-test scalar-labs/scalar-admin-for-kubernetes \
    -f /tmp/test2b-values.yaml -n default

# Confirm pause has started by checking ScalarDB logs                      
kubectl logs -l app.kubernetes.io/app=scalardb-cluster -n default --since=120s | grep -i "Paused" 

# Get first pod name and restart its container (this will increase restart count)
POD_NAME=$(kubectl get pods -l app.kubernetes.io/app=scalardb-cluster -n default -o jsonpath='{.items[0].metadata.name}')
echo "Restarting container in pod: $POD_NAME"
kubectl exec $POD_NAME -n default -- kill -15 1

# Wait for the container to restart and become Ready again
sleep 5
kubectl wait --for=condition=Ready pod $POD_NAME -n default --timeout=60s

# Wait for job to fail
kubectl wait --for=condition=failed job -l app.kubernetes.io/app=scalar-admin-for-kubernetes,app.kubernetes.io/instance=scalar-admin-test \
    -n default --timeout=90s

# Check job logs - should show StatusUnmatchedException as primary error
JOB_POD=$(kubectl get pod -l app.kubernetes.io/app=scalar-admin-for-kubernetes,app.kubernetes.io/instance=scalar-admin-test -n default -o jsonpath='{.items[0].metadata.name}')
kubectl logs $JOB_POD -n default 2>&1 | grep StatusUnmatchedException

# Expected output:
# com.scalar.admin.kubernetes.domain.exception.StatusUnmatchedException: The target pods were updated during the pause duration. You cannot use the backup that was taken during this pause duration.

# Cleanup
helm uninstall scalar-admin-test -n default
```

**Test 3: Deployment annotation change during pause (error case)**

Changing a deployment annotation only affects the deployment's `resourceVersion`, not the pods. Therefore, the unpause operation succeeds and `StatusUnmatchedException` is the primary error. 

```sh
# Deploy with longer pause duration
cat > /tmp/test3-values.yaml <<EOF
scalarAdminForKubernetes:
  image:
    repository: scalar-admin-for-kubernetes-cli
    tag: local-test
    pullPolicy: Never
  jobType: "job"
  commandArgs:
    - --release-name
    - scalardb-cluster
    - --namespace
    - default
    - --pause-duration
    - "60000"  # 60 seconds
EOF

helm install scalar-admin-test scalar-labs/scalar-admin-for-kubernetes \
    -f /tmp/test3-values.yaml -n default

# Confirm pause has started by checking ScalarDB logs                                                                                                
kubectl logs -l app.kubernetes.io/app=scalardb-cluster -n default --since=120s | grep -i "Paused" 

# Change deployment annotation (changes resourceVersion)                                                                                             
kubectl annotate deployment scalardb-cluster-node e2e-test-annotation=test -n default 

# Wait for job to fail
kubectl wait --for=condition=failed job -l app.kubernetes.io/app=scalar-admin-for-kubernetes,app.kubernetes.io/instance=scalar-admin-test \
    -n default --timeout=90s

# Check job logs - should show StatusUnmatchedException as primary error
JOB_POD=$(kubectl get pod -l app.kubernetes.io/app=scalar-admin-for-kubernetes,app.kubernetes.io/instance=scalar-admin-test -n default -o jsonpath='{.items[0].metadata.name}')
kubectl logs $JOB_POD -n default 2>&1 | grep StatusUnmatchedException

# Expected output:
# com.scalar.admin.kubernetes.domain.exception.StatusUnmatchedException: The target pods were updated during the pause duration. You cannot use the backup that was taken during this pause duration.

# Cleanup
helm uninstall scalar-admin-test -n default
kubectl annotate deployment scalardb-cluster-node e2e-test-annotation- -n default
```


lib/src/main/java/com/scalar/admin/kubernetes/domain/model/pause/PauseTarget.java
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Compact constructor with immutability enforcement.
+   *
+   * @param pods the list of pods that are part of this pause target
+   * @param deployment the deployment associated with this pause target
+   * @param adminPort the admin port number used for pause operations
+   */
+  public PauseTarget {
+    pods = ImmutableList.copyOf(pods);
+  }
```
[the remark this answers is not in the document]
monkey-mas: fixed with tests: https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/46/commits/cde0f32b8b40ef80d7640b6e95b0772d7ac22624

lib/src/main/java/com/scalar/admin/kubernetes/domain/model/pause/PauseTarget.java
```diff
@@ -0,0 +1,90 @@
...
+
+    for (V1Pod pod : pods) {
+      String podName = pod.getMetadata().getName();
+      Integer restartCount =
+          pod.getStatus().getContainerStatuses().stream().mapToInt(c -> c.getRestartCount()).sum();
+      String resourceVersion = pod.getMetadata().getResourceVersion();
+
+      podRestartCounts.put(podName, restartCount);
+      podResourceVersions.put(podName, resourceVersion);
+    }
```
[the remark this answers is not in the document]
monkey-mas: fixed with tests; https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/46/changes/daf82cd68ab59137c7eb7b033480892ced5535eb

lib/src/main/java/com/scalar/admin/kubernetes/presentation/dto/PauseRequest.java
```diff
@@ -0,0 +1,46 @@
...
+      throw new IllegalArgumentException("namespace is required");
+    }
+    if (helmReleaseName == null || helmReleaseName.isBlank()) {
+      throw new IllegalArgumentException("helmReleaseName is required");
+    }
+    if (pauseDuration < 1) {
+      throw new IllegalArgumentException(
+          "pauseDuration must be greater than 0, but was: " + pauseDuration);
+    }
+  }
```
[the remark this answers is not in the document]
monkey-mas: fixed with tests: https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/46/changes/ffecf5b12f921eb2f8e9b69a564cf1f509205294

lib/src/main/java/com/scalar/admin/kubernetes/domain/service/PauseService.java
```diff
@@ -66,85 +67,55 @@ public class Pauser {
...
+      PauseTarget targetBeforePause,
+      PauseTargetSupplier targetAfterPauseSupplier,
+      ScalarAdminClient client,
+      int pauseDuration,
+      @Nullable Long maxPauseWaitTime)
       throws PauserException {
     if (pauseDuration < 1) {
       throw new IllegalArgumentException(
           "pauseDuration is required to be greater than 0 millisecond.");
     }
```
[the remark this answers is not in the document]
monkey-mas: fixed with tests: https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/46/changes/aee279ae30c0cb3e6619f72ebb8e37f04ddb8f22. In this case, we can simply use `Objects.requireNonNull` to follow Java convension.

lib/src/main/java/com/scalar/admin/kubernetes/domain/model/shared/Product.java
```diff
@@ -15,21 +15,21 @@ private Product(String appLabelValue, String adminPortName) {
...
-    return UNKNOWN;
+  public static Product fromAppLabelValue(String appLabelValue) {
+    return switch (appLabelValue) {
+      case "scalardb" -> SCALARDB_SERVER;
+      case "scalardb-cluster" -> SCALARDB_CLUSTER;
+      case "ledger" -> SCALARDL_LEDGER;
+      case "auditor" -> SCALARDL_AUDITOR;
+      default -> UNKNOWN;
+    };
   }
```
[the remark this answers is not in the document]
monkey-mas: I don't think a null check is needed here. The switch expression is the very first statement in the method, so a null argument would immediately throw a `NullPointerException` with a clear stack trace.

Additionally, returning `UNKNOWN` for null (as suggested) would silently mask an error. `UNKNOWN` represents a valid label value that doesn't match any known product (e.g., an Envoy pod), whereas null indicates the caller failed to validate its input.

lib/src/main/java/com/scalar/admin/kubernetes/application/PauseApplicationService.java
```diff
@@ -0,0 +1,111 @@
...
+   *
+   * @param command the pause command specifying the operation details
+   * @return DTO containing the start and end time of the pause operation
+   * @throws PauserException when the pause operation fails
+   */
+  public PauseDurationDto execute(PauseCommand command) throws PauserException {
+    return switch (command) {
+      case PauseByHelmReleaseCommand cmd -> executePauseByHelmRelease(cmd);
+    };
+  }
```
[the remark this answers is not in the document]
monkey-mas: @Copilot Similar to the feedback [1], the switch expression is the first statement in the method, so a null command would immediately throw a `NullPointerException` at that point. Adding an explicit null check doesn't improve fail-fast behavior or diagnostics, i.e., the failure location and timing are identical either way.

This is different from the case in [2], where null parameters would cause failures deep inside complex control flow. Here, the failure is immediate and self-evident.

[1] https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/46#discussion_r3332396519
[2] https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/46#discussion_r3332396512

lib/src/main/java/com/scalar/admin/kubernetes/domain/model/shared/Product.java
```diff
@@ -1,6 +1,6 @@
-package com.scalar.admin.kubernetes;
+package com.scalar.admin.kubernetes.domain.model.shared;
 
-enum Product {
+public enum Product {
   SCALARDB_SERVER("scalardb", "scalardb"),
```
feeblefakie: I don't think ScalarDB Server is needed anymore.

lib/src/test/java/com/scalar/admin/kubernetes/application/PauseApplicationServiceTest.java
```diff
@@ -0,0 +1,193 @@
...
+  class Constructor {
+    @Test
+    void constructor_WithValidArgs_CreateInstance() {
+      // Act & Assert
+      assertDoesNotThrow(
+          () -> new PauseApplicationService(kubernetesClient, scalarAdminClientFactory, pauseService));
+    }
+
+    @Test
+    void constructor_WithNullRepository_ThrowIllegalArgumentException() {
```
kota2and3kan:
```suggestion
    void constructor_WithNullKubernetesClient_ThrowIllegalArgumentException() {
```
It seems that the `pauseTargetRepository` was refactored to `kubernetesClient`.

Therefore, it might be better to rename the test name here.

What do you think?

lib/src/main/java/com/scalar/admin/kubernetes/application/PauseApplicationService.java
```diff
@@ -0,0 +1,111 @@
...
+import com.scalar.admin.kubernetes.application.dto.PauseDurationDto;
+import com.scalar.admin.kubernetes.domain.client.ScalarAdminClient;
+import com.scalar.admin.kubernetes.domain.exception.PauserException;
+import com.scalar.admin.kubernetes.domain.model.pause.PauseByHelmReleaseCommand;
+import com.scalar.admin.kubernetes.domain.model.pause.PauseCommand;
+import com.scalar.admin.kubernetes.domain.model.pause.PauseDuration;
+import com.scalar.admin.kubernetes.domain.model.pause.PauseTarget;
+import com.scalar.admin.kubernetes.domain.client.KubernetesClient;
+import com.scalar.admin.kubernetes.domain.service.PauseService;
+import com.scalar.admin.kubernetes.infrastructure.client.ScalarAdminClientFactory;
```
kota2and3kan: Thank you for the discussion in the meeting!

As we discussed, it seems that `PauseApplicationService (application layer)` depends on `ScalarAdminClientFactory (infrastructure layer)` here.

Honestly, I'm not fully sure about the onion architecture, but it might be better to depend on the domain layer instead of that. Actually, all other dependencies' direction seems to be to the domain layer in `PauseApplicationService`.

What do you think?

kota2and3kan (commented):
Thank you for the updates!
I left two suggestions and comments.
Please take a look when you have time! 🙇

lib/src/main/java/com/scalar/admin/kubernetes/application/PauseApplicationService.java
monkey-mas: Agreed - thanks for pointing that out :) 
https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/46/commits/fc11f3c1d7d9d17eccac9cf72dfe9815376c17c0

lib/src/main/java/com/scalar/admin/kubernetes/domain/model/pause/PauseTarget.java
```diff
@@ -0,0 +1,119 @@
...
+                .sum();
+      } else {
+        logger.info(
+            "Pod {} has no status or container statuses; treating restart count as 0.", podName);
+      }
+      String resourceVersion = pod.getMetadata().getResourceVersion();
+
+      podRestartCounts.put(podName, restartCount);
+      podResourceVersions.put(podName, resourceVersion);
+    }
```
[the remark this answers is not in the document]
monkey-mas: You're right that `ImmutableMap.copyOf()` will throw NPE on null values, which is a valid concern. That said, since this PR is focused on refactoring and we want to avoid behavioral changes, we'll address this in a follow-up.

Note:
In practice, `resourceVersion` is always present on objects returned by the Kubernetes API (the pods and deployment here are fetched via `listNamespacedPod/listNamespacedDeployment`). So this is a purely defensive measure.

lib/src/main/java/com/scalar/admin/kubernetes/domain/model/pause/TlsConfig.java
```diff
@@ -0,0 +1,29 @@
...
+   * @throws IllegalArgumentException if either parameter is null
+   */
+  public TlsConfig {
+    if (caRootCert == null) {
+      throw new IllegalArgumentException("caRootCert is required for TLS configuration");
+    }
+    if (overrideAuthority == null) {
+      throw new IllegalArgumentException("overrideAuthority is required for TLS configuration");
+    }
+  }
```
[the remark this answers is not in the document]
monkey-mas: fixed: [28d8aa6](https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/46/commits/28d8aa61cb9053b82d48d2bd1552c67b3cac0dd6)

lib/src/main/java/com/scalar/admin/kubernetes/domain/model/pause/PauseTarget.java
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * <p>This aggregate root encapsulates all information about the target of a pause operation,
+ * including the pods, deployment, and admin port. It provides methods to extract status information
+ * for comparison purposes.
+ *
+ * @param pods the list of pods that are part of this pause target
+ * @param deployment the deployment associated with this pause target
+ * @param adminPort the admin port number used for pause operations
+ */
+public record PauseTarget(List<V1Pod> pods, V1Deployment deployment, int adminPort) {
```
feeblefakie, line 29: Nice refactor.
One layering leak: the aggregate root holds V1Pod/V1Deployment directly, so the domain (and the KubernetesClient/ScalarAdminClient interfaces that return it) can't compile or be tested without the k8s client lib, and the k8s traversal in toStatus() lives in the domain.

Two options:
1. Add a small domain model (e.g. PausePod(name, ip, restartCount, resourceVersion)) and translate from V1Pod/V1Deployment in KubernetesClientImpl.
2. Keep the SDK types as a deliberate choice and add a brief note saying so.

Either's fine. Not blocking.

lib/src/main/java/com/scalar/admin/kubernetes/application/PauseApplicationService.java
```diff
@@ -0,0 +1,111 @@
...
+ * <ol>
+ *   <li>Retrieving the pause target from the repository based on the command
+ *   <li>Creating the appropriate Scalar Admin client (with or without TLS)
+ *   <li>Delegating to the domain service for business logic execution
+ * </ol>
+ *
+ * <p>This class is not thread-safe because it causes side effects in the states of target pods.
+ */
+@NotThreadSafe
+public class PauseApplicationService {
```
feeblefakie, line 30: Inversion is clean everywhere else, since KubernetesClient/ScalarAdminClient are domain interfaces. The one exception is ScalarAdminClientFactory, a concrete infra class held in the application layer, so the dependency arrow points outward to infra here. It's also what makes PauseApplicationServiceTest work around the factory instead of injecting a fake.

Suggestion: extract a ScalarAdminClientFactory interface (in the domain), have infra implement it, and inject that. That closes the last outward edge. Not blocking.

lib/src/main/java/com/scalar/admin/kubernetes/infrastructure/client/ScalarAdminClientImpl.java
```diff
@@ -0,0 +1,56 @@
...
+   *
+   * @param target the pause target
+   * @return list of addresses (pod IP + admin port)
+   */
+  private List<InetSocketAddress> buildAddressList(PauseTarget target) {
+    return target.pods().stream()
+        .map(pod -> new InetSocketAddress(pod.getStatus().getPodIP(), target.adminPort()))
+        .collect(Collectors.toList());
+  }
+}
```
feeblefakie: It seems like buildAddressList calls pod.getStatus().getPodIP() with no guard.
A pod that's gracefully skipped in the status check will NPE when building the client.
Also worth noting that new InetSocketAddress(null, port) doesn't throw; it binds the wildcard address, so a null pod IP could produce a wrong address rather than a clean failure.

Suggestion: filter to pods with a reachable IP and fail loudly if none remain.

```
  List<InetSocketAddress> addrs = target.pods().stream()
      .filter(p -> p.getStatus() != null && p.getStatus().getPodIP() != null)
      .map(p -> new InetSocketAddress(p.getStatus().getPodIP(), target.adminPort()))
      .toList();
  if (addrs.isEmpty()) {
    throw new IllegalStateException("No target pods have a reachable IP.");
  }
```

  Since this helper is duplicated across both client impls, it's also a good candidate to hoist into a shared place. Not blocking.

lib/src/main/java/com/scalar/admin/kubernetes/domain/model/pause/PauseTarget.java
```diff
@@ -0,0 +1,116 @@
...
+            "Pod {} has no status or container statuses; treating restart count as 0.", podName);
+      }
+      String resourceVersion = pod.getMetadata().getResourceVersion();
+
+      podRestartCounts.put(podName, restartCount);
+      podResourceVersions.put(podName, resourceVersion);
+    }
+
+    String deploymentResourceVersion =
+        deployment.getMetadata() != null ? deployment.getMetadata().getResourceVersion() : null;
```
feeblefakie: The added null-safety in toStatus() is reasonable, but it might quietly change the safety contract. Previously, a pod with null metadata/status, or a null deployment resourceVersion, would NPE and surface as StatusCheckFailedException ("you can't trust this backup, retry"). 

Now:
  - a null-metadata pod is dropped from both before and after snapshots, so its churn can't be detected;
  - a pod with no container statuses is recorded as restartCount = 0;
  - two null deployment resourceVersions compare equal (the old hand-written equals would have NPE'd).

So pause() can now report "targets unchanged" from an incomplete snapshot. For a tool whose job is asserting nothing changed during the backup window, I'd lean fail-closed: if a target pod can't be fully read, treat it as "can't verify" and throw StatusCheckFailedException.
If the leniency is intentional, a comment explaining why incomplete data is safe here would help, plus a test pinning the behavior.

lib/src/test/java/com/scalar/admin/kubernetes/domain/service/PauseServiceTest.java
```diff
@@ -1,114 +1,66 @@
...
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
-import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.mockito.MockedStatic;
 
-class PauserTest {
+class PauseServiceTest {
```
feeblefakie, line 45: Three end-to-end combined-failure tests through pause() seemed to be dropped and not re-added:

- pause_WhenUnpauseFailedAndTargetPodStatusChanged_...
- pause_WhenPauseAndUnpauseAndTargetPodStatusChanged_...
- pause_WhenPauseFailedAndTargetPodStatusChanged_...

Only pause_WhenPauseAndUnpauseThrowException survived.
The buildException unit tests prove the ranking is correct in isolation, but nothing now verifies pause() feeds it the right exceptions when multiple operations fail together, which is exactly the primary/suppressed behavior the PR description highlights.

Could you restore equivalent pause()-level tests, ideally also asserting getSuppressed() contents?

lib/src/main/java/com/scalar/admin/kubernetes/infrastructure/client/KubernetesClientImpl.java
feeblefakie, line 176: Nit:
```suggestion
      throw new PauserException(m, e);
```

feeblefakie, line 184: Nit:
```suggestion
      throw new PauserException(m, e);
```

lib/src/main/java/com/scalar/admin/kubernetes/presentation/dto/PauseRequest.java
```diff
@@ -0,0 +1,55 @@
...
+    if (namespace == null || namespace.isBlank()) {
+      throw new IllegalArgumentException("namespace is required");
+    }
+    if (helmReleaseName == null || helmReleaseName.isBlank()) {
+      throw new IllegalArgumentException("helmReleaseName is required");
+    }
+    if (pauseDuration < 1) {
+      throw new IllegalArgumentException(
+          "pauseDuration must be greater than 0, but was: " + pauseDuration);
+    }
```
feeblefakie, line 44: The namespace/release/pauseDuration rules are enforced in three records along a single linear path (PauseRequest to PauseByHelmReleaseCommand to PauseService), where the later checks are effectively unreachable because the input was already validated at the boundary.

Maybe, it's better to validate business rules once at the public boundary (PauseRequest), and keep only Objects.requireNonNull guards deeper down where a null would otherwise NPE (which PauseService.pause() already does well).

lib/src/main/java/com/scalar/admin/kubernetes/domain/model/shared/Product.java
monkey-mas: `SCALARDB_SERVER` is actively used - `fromAppLabelValue("scalardb")` maps to it, and removing it would cause  pods with the `app.kubernetes.io/app=scalardb` label to be treated as `UNKNOWN` and excluded from pause targets. That said, the product naming may be outdated. I'll confirm with @kota2and3kan whether the naming should be updated in a follow-up.

lib/src/main/java/com/scalar/admin/kubernetes/domain/model/pause/PauseTarget.java
monkey-mas: Agreed. This is a deliberate choice for now - I've added a Javadoc note explaining this. A follow-up PR will introduce domain models (e.g., `PausePod(name, ip, restartCount, resourceVersion)`) to decouple from the SDK and enable validation at construction time.

lib/src/main/java/com/scalar/admin/kubernetes/application/PauseApplicationService.java
monkey-mas: Thanks for pointing it out! This has been addressed in a recent commit. https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/46#discussion_r3377481942

`ScalarAdminClientFactory` is now an interface in `domain/client/`, with `ScalarAdminClientFactoryImpl` in `infrastructure/client/`. This closes the last outward dependency from the application layer, consistent with the `KubernetesClient` / `KubernetesClientImpl` pattern.

lib/src/main/java/com/scalar/admin/kubernetes/infrastructure/client/ScalarAdminClientImpl.java
monkey-mas: Agreed. fix: https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/46/commits/e27a5fa89988fa94948db5dfafb5dd8f0b7e8ecd

Added `null` guard (filter out pods with null status or null pod IP, fail loudly if none remain)  and hoisted the duplicated method into `PauseTarget.toAddressList()` to eliminate the duplication across both client impls. Shared utility extraction into a separate class is left for a follow-up (will do in the next PR where I'll introduce another API).

lib/src/main/java/com/scalar/admin/kubernetes/domain/model/pause/PauseTarget.java
monkey-mas: Thanks! fix: https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/46/changes/0d5ac1575b55fde8469b4c3f551fe4c437827b86

I've reverted the null-safety guards in `toStatus()` to preserve the original fail-closed behavior, i.e., if a pod or deployment has incomplete metadata/status, the resulting NPE surfaces as  `StatusCheckFailedException`, which correctly signals "can't verify this backup." Tests pinning this NPE behavior have been added.

I'd say the root cause is that `PauseTarget` holds K8s SDK types (`V1Pod`, `V1Deployment`) directly, so there's no validation at construction time. A follow-up PR will introduce domain models (e.g., `PausePod(name, ip, restartCount, resourceVersion)`) translated in `KubernetesClientImpl`, which will eliminate these NPE paths by validating at the boundary.

lib/src/test/java/com/scalar/admin/kubernetes/domain/service/PauseServiceTest.java
monkey-mas: :saluting_face:
fix: https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/46/commits/332924e5c17cbe108c0fb1842e6635920575080b

lib/src/main/java/com/scalar/admin/kubernetes/infrastructure/client/KubernetesClientImpl.java
monkey-mas: I think we don't have any exceptions thrown here but we simply throw an exception due to the illegal state - but there's another place where we should propagate an exception as you can see.

fix: https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/46/commits/8929f622f17aaa46fb22d5cdc50599a3e03a99dc

monkey-mas: ditto: https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/46/changes/8929f622f17aaa46fb22d5cdc50599a3e03a99dc#r3403023899

lib/src/main/java/com/scalar/admin/kubernetes/presentation/dto/PauseRequest.java
monkey-mas: You're right that the current validation is effectively duplicated along the linear path. This is a consequence of the business rules being simple right now (in the end, the checks happen to be identical...).

Ideally, each layer in DDD should have its own validation, e.g.,                                         
- Presentation boundary (`PauseRequest`): validates input format (non-null, non-blank, positive)
- Domain command (`PauseByHelmReleaseCommand`): protects its own invariants - currently the same checks, but these could diverge                                                                                    
  
For example, [K8s namespaces must be valid DNS-1123 labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#namespaces-and-dns) (`[a-z0-9]([-a-z0-9]*[a-z0-9])?`, max 63 chars), and Helm release names follow a similar but distinct rule (`[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?`)*, max 53 chars). These are domain-level invariants that should be validated in `PauseByHelmReleaseCommand`, not at the presentation boundary.

ref. [kubernetes: namespace validation](https://github.com/kubernetes/kubernetes/blob/v1.36.1/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L155) 
ref. [Helm: release name validation](https://github.com/helm/helm/blob/v4.2.1/pkg/chart/v2/util/validate_name.go#L35)

Let me please keep the command-level validation for now. Adding these rules in this refactoring PR would change behavior, so I'll address it in a follow-up. A `Namespace` value object would be the DDD-ideal approach, but may be overkill at this point - adding the validation directly to the command constructor is likely sufficient.

lib/src/main/java/com/scalar/admin/kubernetes/domain/model/shared/Product.java
feeblefakie: @kota2and3kan Do we still need to keep `scalardb` here?

feeblefakie (approved):
LGTM! Thank you!
We can discuss whether or not keeping `scalardb` later.

lib/src/main/java/com/scalar/admin/kubernetes/domain/model/shared/Product.java
kota2and3kan: @monkey-mas @feeblefakie 
Oh, sorry. I overlooked this comment.
I think we can remove `scalardb` since we don't support ScalarDB Server now.

lib/src/main/java/com/scalar/admin/kubernetes/presentation/PauseController.java
```diff
@@ -0,0 +1,75 @@
...
+    } catch (IOException e) {
+      throw new PauserException("Failed to set default Kubernetes client.", e);
+    }
+
+    KubernetesClient kubernetesClient =
+        new KubernetesClientImpl(new CoreV1Api(), new AppsV1Api());
+    ScalarAdminClientFactoryImpl clientFactory = new ScalarAdminClientFactoryImpl();
+    PauseService pauseService = new PauseService();
+
+    this.applicationService = new PauseApplicationService(kubernetesClient, clientFactory, pauseService);
```
brfrn169: `PauseController` belongs to the presentation layer, but its constructor directly instantiates infrastructure implementations (`KubernetesClientImpl`, `ScalarAdminClientFactoryImpl`) and performs live `Config.defaultClient()` setup. As a result, the presentation layer depends on (and imports, see lines 9-10) the infrastructure layer, which goes against the inward-only dependency direction this PR's onion architecture is meant to establish (`presentation -> application -> domain <- infrastructure`).

It also means the controller can only be constructed against a real Kubernetes environment, so it can't be instantiated/unit-tested in isolation — which seems to be why there is no test covering this class. Flagging this for discussion.

lib/src/main/java/com/scalar/admin/kubernetes/presentation/dto/PauseRequest.java
```diff
@@ -0,0 +1,55 @@
...
+    if (tlsEnabled) {
+      if (caRootCert == null || caRootCert.isBlank()) {
+        throw new IllegalArgumentException("caRootCert is required when tlsEnabled is true");
+      }
+      if (overrideAuthority == null || overrideAuthority.isBlank()) {
+        throw new IllegalArgumentException(
+            "overrideAuthority is required when tlsEnabled is true");
+      }
```
brfrn169, line 52: When `--tls` is enabled, this constructor now rejects null/blank `caRootCert` and `overrideAuthority`, so both effectively become required whenever TLS is used (the previous `TlsPauser` accepted them as nullable). This change isn't reflected in the user-facing documentation: the CLI `--help` text for `--ca-root-cert-path`, `--ca-root-cert-pem`, and `--override-authority` still describes them only as options for use "when wire encryption is enabled", not as required together with `--tls`. The `--help` descriptions　should be updated so the now-mandatory combination is clear to users.

lib/src/main/java/com/scalar/admin/kubernetes/application/PauseApplicationService.java
```diff
@@ -0,0 +1,111 @@
...
+
+  private PauseDurationDto executePauseByHelmRelease(PauseByHelmReleaseCommand command)
+      throws PauserException {
+    // Get the pause target before pause
+    PauseTarget targetBeforePause;
+    try {
+      targetBeforePause =
+          kubernetesClient.resolvePauseTargetByHelmRelease(command.namespace(), command.helmReleaseName());
+    } catch (Exception e) {
+      throw new PauserException("Failed to find the target pods to pause.", e);
```
brfrn169: This wraps whatever `resolvePauseTargetByHelmRelease` throws into a `PauserException("Failed to find the target pods to pause.")`. But `KubernetesClientImpl.resolvePauseTargetByHelmRelease` already throws a `PauserException` with a more specific message (e.g. "Can not find any target pods." / "Helm release <name> didn't create any pod."). The result is a `PauserException` whose cause is another `PauserException`, so an operator sees the generic outer message first and the specific reason is pushed one level deeper in the cause chain. Flagging the double-wrapping of an already-typed domain exception for discussion.

brfrn169 (commented):
Left a few comments. PTAL!

lib/src/main/java/com/scalar/admin/kubernetes/domain/model/shared/Product.java
monkey-mas: removed! :heavy_check_mark:  : https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/46/commits/c3b9cb17a3e781857b75acccb81d9eaaa726854b

lib/src/main/java/com/scalar/admin/kubernetes/presentation/PauseController.java
monkey-mas: Agreed. I introduced Guice (5.1.0) to resolve this. With the Guice refactoring, `PauseController` now accepts `PauseApplicationService` via constructor injection (`@Inject` + a new `PauseModule` in the infrastructure layer handles the wiring), so it's fully unit-testable in isolation. This eliminates all infrastructure imports from the presentation layer.

It felt slightly over-engineered given the current dependency graph size, but I think it's the right call to stay consistent with onion architecture principles. :+1: The `ScalarAdminClientFactory` interface is retained as-is since it takes runtime parameters (`PauseTarget`, `TlsConfig`) that DI alone can't handle.

Let me know if you'd prefer a different approach. ^

fix with DI: fixed in https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/46/commits/9da707e54ea43252d8d5fae08d39b59bb9e96ce7

lib/src/main/java/com/scalar/admin/kubernetes/presentation/dto/PauseRequest.java
monkey-mas: Thanks for the feedback! Fixed the wording so that users can figure out that these parameters are actually required.
description updates: https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/46/commits/b98c97a87a1fe7c42b1d98dd1317fe7b33ff3046

lib/src/main/java/com/scalar/admin/kubernetes/application/PauseApplicationService.java
monkey-mas: Agreed. Removed the try-catch around `resolvePauseTargetByHelmRelease()` entirely so the infrastructure layer's specific `PauserException` messages propagate directly without double-wrapping.

I kept the try-catch around clientFactory.createClient() as-is because `createClient()` doesn't declare throws `PauserException` (it's a simple object instantiation), so double-wrapping can't happen thereonly (i.e., unexpected `RuntimeExceptions` would be caught).

brfrn169 (approved):
LGTM! Thank you!

PR #47, opened 2026-06-01
Introduce `--pod-discovery-mode` to support (Helm-free) pausing by Deployment name

monkey-mas:
## Description

This PR adds support for pausing Scalar products by **Kubernetes Deployment name** in addition to the existing Helm release name approach. This feature is built on top of the Onion Architecture refactoring completed in #46.

Currently, `scalar-admin-for-kubernetes` only supports pausing Scalar products by specifying a Helm release name (`--release-name`). This creates limitations in the following scenarios:

1. **Non-Helm deployments**: Deployments created with `kubectl apply` or other tools cannot be paused
2. **Helm dependency**: Users must have Helm installed and access to release metadata
3. **CI/CD environments**: Automated environments may have direct access to deployment names but not Helm release information

By introducing a deployment mode, this tool becomes more versatile and easier to integrate into various operational workflows, removing the hard dependency on Helm for pause operations.

## Related issues and/or PRs

- Depends on #46 (Onion Architecture refactoring)

### 1. New CLI Options

Added three new command-line options to support deployment mode:

- `--pod-discovery-mode` (default: `helm-release` for **backward compatibility**)
  - Values: `helm-release` | `deployment`
  - Specifies how to discover target pods

- `--deployment-name` (required when `--pod-discovery-mode=deployment`)
  - The name of the Kubernetes Deployment for the Scalar product

- `--admin-port` (required when `--pod-discovery-mode=deployment`)
  - The port number of the admin interface (e.g.,60053)

#### Usage Examples

**Existing Helm Release Mode (unchanged):**
```bash
java -jar scalar-admin-for-kubernetes-cli.jar \
  --release-name scalardb-cluster \
  --namespace default \
  --pause-duration 5000
```

**New Helm Release Mode:**
```bash
java -jar scalar-admin-for-kubernetes-cli.jar \
  --pod-discovery-mode helm-release \
  --release-name scalardb-cluster \
  --namespace default \
  --pause-duration 5000
```

**New Deployment Mode:**
```bash
java -jar scalar-admin-for-kubernetes-cli.jar \
  --pod-discovery-mode deployment \
  --deployment-name scalardb-cluster-node \
  --admin-port 60053 \
  --namespace default \
  --pause-duration 5000
```

### 2. Architecture Implementation (7 Phases)

Following Onion Architecture principles:

**Phase 1 - Domain Layer**: Introduced `PodDiscoveryMode` enum
- `lib/src/main/java/com/scalar/admin/kubernetes/domain/model/pause/PodDiscoveryMode.java`
- Core domain concept with validation logic

**Phase 2 - Presentation Layer**: Added CLI option parsing
- `cli/src/main/java/com/scalar/admin/kubernetes/Cli.java`
- Case-insensitive enum value parsing

**Phase 3 - Domain Layer**: Created `PauseByDeploymentNameCommand`
- `lib/src/main/java/com/scalar/admin/kubernetes/domain/model/pause/PauseByDeploymentNameCommand.java`
- New domain command implementing `PauseCommand` interface

**Phase 4 - Infrastructure Layer**: Implemented deployment-based pod discovery
- `lib/src/main/java/com/scalar/admin/kubernetes/domain/client/KubernetesClient.java`
- `lib/src/main/java/com/scalar/admin/kubernetes/infrastructure/client/KubernetesClientImpl.java`
- Added `resolvePauseTargetByDeploymentName()` method using Kubernetes label selectors

**Phase 5 - Application Layer**: Updated `PauseApplicationService`
- `lib/src/main/java/com/scalar/admin/kubernetes/application/PauseApplicationService.java`
- Pattern matching on command type to route to appropriate method

**Phase 6 - Presentation Layer**: Updated `PauseController`
- `lib/src/main/java/com/scalar/admin/kubernetes/presentation/PauseController.java`
- Creates appropriate command based on `PodDiscoveryMode`

**Phase 7 - CLI Output**: Updated `Result` class
- `cli/src/main/java/com/scalar/admin/kubernetes/Result.java`
- Supports both modes in JSON output with `pod_discovery_mode`, `helm_release_name`, and `deployment_name` fields

### 3. E2E Testing

**Enhanced existing E2E test** (`.github/workflows/e2e-helm-release.yml`):
- Renamed from `e2e.yml` to `e2e-helm-release.yml` for clarity
- Added matrix dimension to test both implicit and explicit pod discovery modes:
  - Without `--pod-discovery-mode` (backward compatibility)
  - With `--pod-discovery-mode helm-release` (explicit mode)

**Added new E2E test** (`.github/workflows/e2e-deployment.yml`):
- Tests deployment mode with ScalarDB Cluster, ScalarDL Ledger, and ScalarDL Auditor
- Uses matrix strategy for multiple products
- Dynamically discovers deployment names using Kubernetes label selectors
- Validates pause/unpause with correct admin ports:
  - ScalarDB Cluster: port 60053
  - ScalarDL Ledger: port 50053
  - ScalarDL Auditor: port 40053

## Checklist

> The following is a best-effort checklist. If any items in this checklist are not applicable to this PR or are dependent on other, unmerged PRs, please still mark the checkboxes after you have read and understood each item.

- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have updated the documentation to reflect the changes.
- [x] Any remaining open issues linked to this PR are documented and up-to-date (Jira, GitHub, etc.).
- [x] Tests (unit, integration, etc.) have been added for the changes.
- [x] My changes generate no new warnings.
- [x] Any dependent changes in other PRs have been merged and published.

### Manual E2E Testing
Assuming ScalarDB Cluster is already deployed on a Kubernetes cluster, set up the test client and run the tool as follows:

1. Deploy the client pod

```sh
kubectl apply -f .github/manifests/client-pod.yaml
kubectl wait --for=condition=Available deployment/foo-scalardb-cluster-node --timeout 30s
kubectl wait --for=condition=Ready pod/java21 --timeout 30s
```

2. Build the Shadow JAR and copy it to the client pod

```sh
./gradlew shadowJar

VERSION=$(./gradlew :cli:properties -q | grep "version:" | awk '{print $2}')

kubectl cp cli/build/libs/scalar-admin-for-kubernetes-cli-${VERSION}.jar java21:/tmp
```

3. Run the tool

Deployment mode (new feature in this PR):
```sh
kubectl exec -it java21 -- java -jar /tmp/scalar-admin-for-kubernetes-cli-${VERSION}.jar \
  --pod-discovery-mode deployment \
  --deployment-name scalardb-cluster-node \
  --admin-port 60053 \
  --namespace default
```

Helm release mode (existing feature, regression check):
```sh
kubectl exec -it java21 -- java -jar /tmp/scalar-admin-for-kubernetes-cli-${VERSION}.jar \
  --release-name scalardb-cluster \
  --namespace default
```

4. Expected output

Deployment mode:
```json
{
  "namespace": "default",
  "timezone": "Etc/UTC",
  "pod_discovery_mode": "deployment",
  "deployment_name": "scalardb-cluster-node",
  "pause_start_timestamp_ms": 1782234096290,
  "pause_end_timestamp_ms": 1782234101290,
  "pause_start_date_time": "2026-06-23T17:01:36.290",
  "pause_end_date_time": "2026-06-23T17:01:41.290"
}
```

Helm release mode:
```json
{
  "namespace": "default",
  "timezone": "Etc/UTC",
  "pod_discovery_mode": "helm-release",
  "helm_release_name": "scalardb-cluster",
  "pause_start_timestamp_ms": 1782234299793,
  "pause_end_timestamp_ms": 1782234304793,
  "pause_start_date_time": "2026-06-23T17:04:59.793",
  "pause_end_date_time": "2026-06-23T17:05:04.793"
}
```

Helm release mode with explicit --pod-discovery-mode:
```json
{
  "namespace": "default",
  "timezone": "Etc/UTC",
  "pod_discovery_mode": "helm-release",
  "helm_release_name": "scalardb-cluster",
  "pause_start_timestamp_ms": 1782234858523,
  "pause_end_timestamp_ms": 1782234863523,
  "pause_start_date_time": "2026-06-23T17:14:18.523",
  "pause_end_date_time": "2026-06-23T17:14:23.523"
}
```

5. clean up
```sh
  kubectl delete -f .github/manifests/client-pod.yaml
```

### DIFF
--- a/cli/src/main/java/com/scalar/admin/kubernetes/Result.java
+++ b/cli/src/main/java/com/scalar/admin/kubernetes/Result.java
@@ -1,9 +1,12 @@
 package com.scalar.admin.kubernetes;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.scalar.admin.kubernetes.application.dto.PauseDurationDto;
+
 import java.time.Instant;
 import java.time.ZoneId;
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
 @Immutable
@@ -11,8 +14,18 @@ class Result {
 
   public final String namespace;
 
+  @JsonProperty("pod_discovery_mode")
+  public final String podDiscoveryMode;
+
   @JsonProperty("helm_release_name")
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  @Nullable
   public final String helmReleaseName;
+
+  @JsonProperty("deployment_name")
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  @Nullable
+  public final String deploymentName;
 
   @JsonProperty("pause_start_timestamp_ms")
   public final long pauseStartTimestampMs;
@@ -29,9 +42,16 @@ class Result {
   public final String timezone;
 
   Result(
-      String namespace, String helmReleaseName, PauseDurationDto pauseDurationDto, ZoneId zoneId) {
+      String namespace,
+      String podDiscoveryMode,
+      @Nullable String helmReleaseName,
+      @Nullable String deploymentName,
+      PauseDurationDto pauseDurationDto,
+      ZoneId zoneId) {
     this.namespace = namespace;
+    this.podDiscoveryMode = podDiscoveryMode;
     this.helmReleaseName = helmReleaseName;
+    this.deploymentName = deploymentName;
     this.pauseStartTimestampMs = pauseDurationDto.startTimeEpochMilli();
     this.pauseEndTimestampMs = pauseDurationDto.endTimeEpochMilli();
     this.pauseStartDateTime =
--- a/lib/src/main/java/com/scalar/admin/kubernetes/application/PauseApplicationService.java
+++ b/lib/src/main/java/com/scalar/admin/kubernetes/application/PauseApplicationService.java
@@ -3,6 +3,7 @@ package com.scalar.admin.kubernetes.application;
 import com.scalar.admin.kubernetes.application.dto.PauseDurationDto;
 import com.scalar.admin.kubernetes.domain.client.ScalarAdminClient;
 import com.scalar.admin.kubernetes.domain.exception.PauserException;
+import com.scalar.admin.kubernetes.domain.model.pause.PauseByDeploymentNameCommand;
 import com.scalar.admin.kubernetes.domain.model.pause.PauseByHelmReleaseCommand;
 import com.scalar.admin.kubernetes.domain.model.pause.PauseCommand;
 import com.scalar.admin.kubernetes.domain.model.pause.PauseDuration;
@@ -69,6 +70,8 @@ public class PauseApplicationService {
   public PauseDurationDto execute(PauseCommand command) throws PauserException {
     return switch (command) {
       case PauseByHelmReleaseCommand cmd -> executePauseByHelmRelease(cmd);
+      case PauseByDeploymentNameCommand cmd -> throw new UnsupportedOperationException(
+          "DEPLOYMENT mode is not yet implemented. Only HELM_RELEASE mode is currently supported.");
     };
   }
 
--- a/lib/src/main/java/com/scalar/admin/kubernetes/application/PauseApplicationService.java
+++ b/lib/src/main/java/com/scalar/admin/kubernetes/application/PauseApplicationService.java
@@ -8,9 +8,12 @@ import com.scalar.admin.kubernetes.domain.model.pause.PauseByHelmReleaseCommand;
 import com.scalar.admin.kubernetes.domain.model.pause.PauseCommand;
 import com.scalar.admin.kubernetes.domain.model.pause.PauseDuration;
 import com.scalar.admin.kubernetes.domain.model.pause.PauseTarget;
+import com.scalar.admin.kubernetes.domain.model.pause.TlsConfig;
 import com.scalar.admin.kubernetes.domain.client.KubernetesClient;
 import com.scalar.admin.kubernetes.domain.service.PauseService;
 import com.scalar.admin.kubernetes.domain.client.ScalarAdminClientFactory;
+import com.scalar.admin.kubernetes.domain.service.PauseService.PauseTargetSupplier;
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.NotThreadSafe;
 import javax.inject.Inject;
 
@@ -66,27 +69,70 @@ public class PauseApplicationService {
    * @param command the pause command specifying the operation details
    * @return DTO containing the start and end time of the pause operation
    * @throws PauserException when the pause operation fails
+   * @throws IllegalArgumentException when the command is null
    */
   public PauseDurationDto execute(PauseCommand command) throws PauserException {
+    if (command == null) {
+      throw new IllegalArgumentException("command is required");
+    }
     return switch (command) {
       case PauseByHelmReleaseCommand cmd -> executePauseByHelmRelease(cmd);
-      case PauseByDeploymentNameCommand cmd -> throw new UnsupportedOperationException(
-          "DEPLOYMENT mode is not yet implemented. Only HELM_RELEASE mode is currently supported.");
+      case PauseByDeploymentNameCommand cmd -> executePauseByDeploymentName(cmd);
     };
   }
 
   private PauseDurationDto executePauseByHelmRelease(PauseByHelmReleaseCommand command)
       throws PauserException {
+    return executePause(
+        () ->
+            kubernetesClient.resolvePauseTargetByHelmRelease(
+                command.namespace(), command.helmReleaseName()),
+        command.tlsConfig(),
+        command.pauseDuration(),
+        command.maxPauseWaitTime());
+  }
+
+  private PauseDurationDto executePauseByDeploymentName(PauseByDeploymentNameCommand command)
+      throws PauserException {
+    return executePause(
+        () ->
+            kubernetesClient.resolvePauseTargetByDeploymentName(
+                command.namespace(), command.deploymentName(), command.adminPort()),
+        command.tlsConfig(),
+        command.pauseDuration(),
+        command.maxPauseWaitTime());
+  }
+
+  /**
+   * Executes a pause operation with the given target supplier and configuration.
+   *
+   * @param targetSupplier supplier function to retrieve the pause target
+   * @param tlsConfig TLS configuration (null for non-TLS)
+   * @param pauseDuration the duration to pause in milliseconds
+   * @param maxPauseWaitTime the maximum wait time for pause operation (null for default)
+   * @return DTO containing the start and end time of the pause operation
+   * @throws PauserException when the pause operation fails
+   */
+  private PauseDurationDto executePause(
+      PauseTargetSupplier targetSupplier,
+      @Nullable TlsConfig tlsConfig,
+      int pauseDuration,
+      @Nullable Long maxPauseWaitTime)
+      throws PauserException {
+
     // Get the pause target before pause
-    PauseTarget targetBeforePause =
-        kubernetesClient.resolvePauseTargetByHelmRelease(
-            command.namespace(), command.helmReleaseName());
+    PauseTarget targetBeforePause;
+    try {
+      targetBeforePause = targetSupplier.get();
+    } catch (Exception e) {
+      throw new PauserException("Failed to find the target pods to pause.", e);
+    }
 
     // Create the appropriate client (with or without TLS)
     ScalarAdminClient client;
     try {
-      if (command.tlsConfig() != null) {
-        client = clientFactory.createClient(targetBeforePause, command.tlsConfig());
+      if (tlsConfig != null) {
+        client = clientFactory.createClient(targetBeforePause, tlsConfig);
       } else {
         client = clientFactory.createClient(targetBeforePause);
       }
@@ -95,18 +141,12 @@ public class PauseApplicationService {
     }
 
     // Execute the pause operation through the domain service
-    PauseDuration pauseDuration =
+    PauseDuration duration =
         pauseService.pause(
-            targetBeforePause,
-            () ->
-                kubernetesClient.resolvePauseTargetByHelmRelease(
-                    command.namespace(), command.helmReleaseName()),
-            client,
-            command.pauseDuration(),
-            command.maxPauseWaitTime());
+            targetBeforePause, targetSupplier, client, pauseDuration, maxPauseWaitTime);
 
     // Convert domain object to DTO
     return new PauseDurationDto(
-        pauseDuration.startTime().toEpochMilli(), pauseDuration.endTime().toEpochMilli());
+        duration.startTime().toEpochMilli(), duration.endTime().toEpochMilli());
   }
 }
--- a/lib/src/main/java/com/scalar/admin/kubernetes/application/PauseApplicationService.java
+++ b/lib/src/main/java/com/scalar/admin/kubernetes/application/PauseApplicationService.java
@@ -124,6 +124,8 @@ public class PauseApplicationService {
     PauseTarget targetBeforePause;
     try {
       targetBeforePause = targetSupplier.get();
+    } catch (PauserException e) {
+      throw e;
     } catch (Exception e) {
       throw new PauserException("Failed to find the target pods to pause.", e);
     }
--- a/lib/src/main/java/com/scalar/admin/kubernetes/domain/model/pause/PauseByDeploymentNameCommand.java
+++ b/lib/src/main/java/com/scalar/admin/kubernetes/domain/model/pause/PauseByDeploymentNameCommand.java
@@ -39,18 +39,16 @@ public record PauseByDeploymentNameCommand(
    */
   public PauseByDeploymentNameCommand {
     if (namespace == null || namespace.isBlank()) {
-      throw new IllegalArgumentException("namespace is required");
+      throw new IllegalArgumentException(NAMESPACE_REQUIRED_ERROR);
     }
     if (deploymentName == null || deploymentName.isBlank()) {
-      throw new IllegalArgumentException("deploymentName is required");
+      throw new IllegalArgumentException(DEPLOYMENT_NAME_REQUIRED_ERROR);
     }
     if (adminPort < 1 || adminPort > 65535) {
-      throw new IllegalArgumentException(
-          "adminPort must be between 1 and 65535, but was: " + adminPort);
+      throw new IllegalArgumentException(String.format(ADMIN_PORT_ERROR, adminPort));
     }
     if (pauseDuration < 1) {
-      throw new IllegalArgumentException(
-          "pauseDuration must be greater than 0 millisecond, but was: " + pauseDuration);
+      throw new IllegalArgumentException(String.format(PAUSE_DURATION_ERROR, pauseDuration));
     }
   }
 
--- a/lib/src/main/java/com/scalar/admin/kubernetes/domain/model/pause/PauseByDeploymentNameCommand.java
+++ b/lib/src/main/java/com/scalar/admin/kubernetes/domain/model/pause/PauseByDeploymentNameCommand.java
@@ -1,0 +1,115 @@
+package com.scalar.admin.kubernetes.domain.model.pause;
+
+import javax.annotation.Nullable;
+
+/**
+ * Command to pause pods identified by deployment name.
+ *
+ * <p>This command represents the use case of pausing all pods that belong to a specific deployment
+ * in a given namespace. The admin port must be explicitly specified as it cannot be auto-detected
+ * from deployment labels.
+ *
+ * @param namespace the Kubernetes namespace where the deployment exists
+ * @param deploymentName the name of the deployment
+ * @param adminPort the admin port number for the Scalar product
+ * @param pauseDuration the duration to pause in milliseconds
+ * @param maxPauseWaitTime the maximum wait time (in milliseconds) for pause operation to complete,
+ *     null for default
+ * @param tlsConfig the TLS configuration for secure communication, null for non-TLS communication
+ */
+public record PauseByDeploymentNameCommand(
+    String namespace,
+    String deploymentName,
+    int adminPort,
+    int pauseDuration,
+    @Nullable Long maxPauseWaitTime,
+    @Nullable TlsConfig tlsConfig)
+    implements PauseCommand {
+
+  /**
+   * Compact constructor with validation.
+   *
+   * @param namespace the Kubernetes namespace (required)
+   * @param deploymentName the deployment name (required)
+   * @param adminPort the admin port number (must be 0-65535)
+   * @param pauseDuration the pause duration in milliseconds (must be positive)
+   * @param maxPauseWaitTime the maximum wait time (optional)
+   * @param tlsConfig the TLS configuration (optional)
+   * @throws IllegalArgumentException if required parameters are null or invalid
+   */
+  public PauseByDeploymentNameCommand {
+    if (namespace == null || namespace.isBlank()) {
+      throw new IllegalArgumentException("namespace is required");
+    }
+    if (deploymentName == null || deploymentName.isBlank()) {
+      throw new IllegalArgumentException("deploymentName is required");
+    }
+    if (adminPort < 0 || adminPort > 65535) {
+      throw new IllegalArgumentException(
+          "adminPort must be between 0 and 65535, but was: " + adminPort);
+    }
+    if (pauseDuration < 1) {
+      throw new IllegalArgumentException(
+          "pauseDuration must be greater than 0 millisecond, but was: " + pauseDuration);
+    }
+  }
+
+  /**
+   * Creates a command for pausing pods without TLS.
+   *
+   * @param namespace the Kubernetes namespace
+   * @param deploymentName the deployment name
+   * @param adminPort the admin port number
+   * @param pauseDuration the pause duration in milliseconds
+   * @param maxPauseWaitTime the maximum wait time in milliseconds
+   * @return a new PauseByDeploymentNameCommand without TLS
+   */
+  public static PauseByDeploymentNameCommand create(
+      String namespace,
+      String deploymentName,
+      int adminPort,
+      int pauseDuration,
+      Long maxPauseWaitTime) {
+    return new PauseByDeploymentNameCommand(
+        namespace, deploymentName, adminPort, pauseDuration, maxPauseWaitTime, null);
+  }
+
+  /**
+   * Creates a command for pausing pods with TLS enabled.
+   *
+   * @param namespace the Kubernetes namespace
+   * @param deploymentName the deployment name
+   * @param adminPort the admin port number
+   * @param pauseDuration the pause duration in milliseconds
+   * @param maxPauseWaitTime the maximum wait time in milliseconds
+   * @param caRootCert the CA root certificate
+   * @param overrideAuthority the override authority for TLS
+   * @return a new PauseByDeploymentNameCommand with TLS configuration
+   */
+  public static PauseByDeploymentNameCommand createWithTls(
+      String namespace,
+      String deploymentName,
+      int adminPort,
+      int pauseDuration,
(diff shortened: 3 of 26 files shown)


lib/src/main/java/com/scalar/admin/kubernetes/domain/model/pause/PauseTarget.java
```diff
@@ -0,0 +1,90 @@
...
+      String podName = pod.getMetadata().getName();
+      Integer restartCount =
+          pod.getStatus().getContainerStatuses().stream().mapToInt(c -> c.getRestartCount()).sum();
+      String resourceVersion = pod.getMetadata().getResourceVersion();
+
+      podRestartCounts.put(podName, restartCount);
+      podResourceVersions.put(podName, resourceVersion);
+    }
+
+    String deploymentResourceVersion = deployment.getMetadata().getResourceVersion();
```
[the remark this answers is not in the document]
monkey-mas: We'll do refactoring for `V1Pod` (from k8s library) in `domain`, which should fix this.
ref. https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/46#discussion_r3393957290

lib/src/main/java/com/scalar/admin/kubernetes/infrastructure/client/ScalarAdminClientImpl.java
```diff
@@ -0,0 +1,56 @@
...
+   * Builds a list of InetSocketAddress from the pause target.
+   *
+   * @param target the pause target
+   * @return list of addresses (pod IP + admin port)
+   */
+  private List<InetSocketAddress> buildAddressList(PauseTarget target) {
+    return target.pods().stream()
+        .map(pod -> new InetSocketAddress(pod.getStatus().getPodIP(), target.adminPort()))
+        .collect(Collectors.toList());
+  }
```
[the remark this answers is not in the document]
monkey-mas: Fixed in the previous refactoring PR: https://github.com/scalar-labs/scalar-admin-for-kubernetes/commit/e27a5fa89988fa94948db5dfafb5dd8f0b7e8ecd

lib/src/main/java/com/scalar/admin/kubernetes/infrastructure/client/KubernetesClientImpl.java
```diff
@@ -51,15 +55,112 @@ TargetSnapshot select() throws PauserException {
...
+          String.format(
+              "Kubernetes readNamespacedDeployment API error with code %d and body %s.",
+              e.getCode(), e.getResponseBody());
+      throw new PauserException(m, e);
+    }
+  }
+
+  String extractLabelSelectorFromDeployment(V1Deployment deployment) throws PauserException {
+    String deploymentName = deployment.getMetadata().getName();
+    String namespace = deployment.getMetadata().getNamespace();
```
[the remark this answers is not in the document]
monkey-mas: ditto: this should be fixed once we refator `V1Deployment`
https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/47#discussion_r3457794873

lib/src/main/java/com/scalar/admin/kubernetes/domain/model/shared/Product.java
```diff
@@ -15,21 +15,21 @@ private Product(String appLabelValue, String adminPortName) {
...
-    return UNKNOWN;
+  public static Product fromAppLabelValue(String appLabelValue) {
+    return switch (appLabelValue) {
+      case "scalardb" -> SCALARDB_SERVER;
+      case "scalardb-cluster" -> SCALARDB_CLUSTER;
+      case "ledger" -> SCALARDL_LEDGER;
+      case "auditor" -> SCALARDL_AUDITOR;
+      default -> UNKNOWN;
+    };
   }
```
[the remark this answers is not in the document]
monkey-mas: `null` is invalid here, thus we should throw a run-time exception instead. eg NullPoiterException.

lib/src/main/java/com/scalar/admin/kubernetes/infrastructure/client/TlsScalarAdminClientImpl.java
```diff
@@ -0,0 +1,63 @@
...
+   * Builds a list of InetSocketAddress from the pause target.
+   *
+   * @param target the pause target
+   * @return list of addresses (pod IP + admin port)
+   */
+  private List<InetSocketAddress> buildAddressList(PauseTarget target) {
+    return target.pods().stream()
+        .map(pod -> new InetSocketAddress(pod.getStatus().getPodIP(), target.adminPort()))
+        .collect(Collectors.toList());
+  }
```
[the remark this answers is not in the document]
monkey-mas: fixed in the previous refactoring PR:
https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/46#discussion_r3394056630

lib/src/main/java/com/scalar/admin/kubernetes/infrastructure/client/KubernetesClientImpl.java
```diff
@@ -51,15 +55,112 @@ TargetSnapshot select() throws PauserException {
...
+      String m =
+          String.format(
+              "Kubernetes listNamespacedPod API error with code %d and body %s.",
+              e.getCode(), e.getResponseBody());
+      throw new PauserException(m, e);
+    }
+
+    List<V1Pod> pods = podList.getItems();
+
+    if (pods.isEmpty()) {
```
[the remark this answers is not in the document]
monkey-mas: ditto: this should be fixed once we refator V1Deployment
https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/47#discussion_r3457794873

cli/src/main/java/com/scalar/admin/kubernetes/Result.java
```diff
@@ -29,9 +38,14 @@ class Result {
   public final String timezone;
 
   Result(
-      String namespace, String helmReleaseName, PauseDurationDto pauseDurationDto, ZoneId zoneId) {
+      String namespace,
+      @Nullable String helmReleaseName,
+      @Nullable String deploymentName,
+      PauseDurationDto pauseDurationDto,
```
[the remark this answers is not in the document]
monkey-mas: fixed in https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/47/commits/52131c49125a44ba4f01f6b1355e2b7c6eb8c698

cli/src/main/java/com/scalar/admin/kubernetes/Result.java
```diff
@@ -12,8 +14,15 @@ class Result {
...
   @JsonProperty("helm_release_name")
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  @Nullable
   public final String helmReleaseName;
 
+  @JsonProperty("deployment_name")
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  @Nullable
+  public final String deploymentName;
+
```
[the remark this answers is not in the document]
monkey-mas: fixed in https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/47/commits/52131c49125a44ba4f01f6b1355e2b7c6eb8c698

lib/src/main/java/com/scalar/admin/kubernetes/domain/model/pause/PauseByDeploymentNameCommand.java
```diff
@@ -0,0 +1,115 @@
...
+    if (namespace == null || namespace.isBlank()) {
+      throw new IllegalArgumentException("namespace is required");
+    }
+    if (deploymentName == null || deploymentName.isBlank()) {
+      throw new IllegalArgumentException("deploymentName is required");
+    }
+    if (adminPort < 0 || adminPort > 65535) {
+      throw new IllegalArgumentException(
+          "adminPort must be between 0 and 65535, but was: " + adminPort);
+    }
```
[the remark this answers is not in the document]
monkey-mas: fixed: https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/47/commits/c3fbf0e0c8daac9af4b506ff0a4fac0973fcad20

lib/src/main/java/com/scalar/admin/kubernetes/application/PauseApplicationService.java
```diff
@@ -0,0 +1,150 @@
...
+      @Nullable Long maxPauseWaitTime)
+      throws PauserException {
+
+    // Get the pause target before pause
+    PauseTarget targetBeforePause;
+    try {
+      targetBeforePause = targetSupplier.get();
+    } catch (Exception e) {
+      throw new PauserException("Failed to find the target pods to pause.", e);
+    }
```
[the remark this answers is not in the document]
monkey-mas: https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/47/commits/25dc8092ad566847b29e17aa539162d779e432a7

lib/src/main/java/com/scalar/admin/kubernetes/infrastructure/client/KubernetesClientImpl.java
```diff
@@ -62,6 +64,103 @@ public PauseTarget resolvePauseTargetByHelmRelease(String namespace, String helm
...
+      String m =
+          String.format(
+              "Kubernetes readNamespacedDeployment API error with code %d and body %s.",
+              e.getCode(), e.getResponseBody());
+      throw new PauserException(m, e);
+    }
+  }
+
+  String extractLabelSelectorFromDeployment(V1Deployment deployment) throws PauserException {
+    String deploymentName = deployment.getMetadata().getName();
```
feeblefakie, line 96: deployment.getMetadata() can be null?

lib/src/test/java/com/scalar/admin/kubernetes/application/PauseApplicationServiceTest.java
feeblefakie, line 29: It seems like it dropped the test that threw a real PauserException and asserted its original message survived; the replacement throws a RuntimeException and asserts the wrapped message. So the exact behavior commit 37747af ("avoid double-wrapping PauseException") added now has zero coverage; a regression that re-wrapped PauserException would pass all current tests.

How about adding a test (deployment and/or helm) where kubernetesClient.resolvePauseTarget… throws a PauserException, asserting it propagates unwrapped; e.g. assertThatThrownBy(...).isSameAs(stub).hasMessage("original")?

lib/src/main/java/com/scalar/admin/kubernetes/infrastructure/client/KubernetesClientImpl.java
monkey-mas: Thank you for catching this. :pray: `V1Deployment.getMetadata()` is indeed `@Nullable` per the Kubernetes Java client API, although in practice it is never null for objects returned by the API since a succeful Kubernetes API response should have deployment data.

Anyways, to be on the safe side, I added a defensive null check with a comment: 69fd3a1

As discussed in https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/46#discussion_r3393957290, a follow-up PR will introduce domain models (e.g., DomainDeployment) to decouple from SDK types (eg V1Deployment). Once that's done, these null checks will no longer be needed since validation will happen at construction time in the infrastructure layer.

lib/src/test/java/com/scalar/admin/kubernetes/application/PauseApplicationServiceTest.java
monkey-mas: Sorry for dropping the ball :facepalm: 

To cover these test cases, I added propagation tests for both helm-release and deployment modes using `isSameAs(original)` to verify object identity (no wrapping). Also renamed the existing `RuntimeException` test for clarity.

fixed: https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/47/commits/2b7216d69f4f2e40ae8f9b3c9a6c6dfe9c3efda5

lib/src/main/java/com/scalar/admin/kubernetes/domain/model/pause/PauseByDeploymentNameCommand.java
```diff
@@ -0,0 +1,115 @@
...
+      throw new IllegalArgumentException("namespace is required");
+    }
+    if (deploymentName == null || deploymentName.isBlank()) {
+      throw new IllegalArgumentException("deploymentName is required");
+    }
+    if (adminPort < 1 || adminPort > 65535) {
+      throw new IllegalArgumentException(
+          "adminPort must be between 1 and 65535, but was: " + adminPort);
+    }
+    if (pauseDuration < 1) {
```
feeblefakie, line 50: **P2 — Validation rules are duplicated across layers with divergent error messages.** The same rules are enforced in `PauseRequest`, `PodDiscoveryMode.validate()`, and the command compact constructors, and the message strings have drifted, so identical bad input yields different errors depending on the entry point:

- `pauseDuration`: this line says `"...greater than 0 millisecond, but was: X"`, but `PauseRequest.java:47-49` says `"...greater than 0, but was: X"` (no "millisecond"), and `PauseService.java:93-94` says `"...required to be greater than 0 millisecond."` — three strings for one rule.
- `adminPort` range: `PodDiscoveryMode.java:101-103` (`"...between 1 and 65535 when podDiscoveryMode is DEPLOYMENT..."`) vs. line 47-49 here (`"...between 1 and 65535..."`).

The CLI path is deterministic (the user sees the outermost message), so the deeper duplicates are effectively unreachable from the CLI and only surface when the records are used directly as a library — but the inconsistency is a maintenance trap (a future rule change must touch up to four places). Suggest: keep the command ctor as the single source of truth for *value* rules, narrow `PodDiscoveryMode.validate()` to the cross-field "which args are allowed/required per mode" checks, and unify shared messages into constants.

lib/src/main/java/com/scalar/admin/kubernetes/domain/model/pause/PodDiscoveryMode.java
```diff
@@ -0,0 +1,112 @@
...
+   * @param helmReleaseName the Helm release name (required for HELM_RELEASE mode)
+   * @param deploymentName the deployment name (required for DEPLOYMENT mode)
+   * @param adminPort the admin port (required for DEPLOYMENT mode)
+   * @throws IllegalArgumentException if required parameters are missing or invalid for this mode
+   */
+  public void validate(
+      @Nullable String helmReleaseName,
+      @Nullable String deploymentName,
+      @Nullable Integer adminPort) {
+    switch (this) {
```
feeblefakie: **P2 — This `switch (this)` has no `default`, so a future third mode silently skips validation.** Unlike the two dispatch sites that are compiler-checked for exhaustiveness (`PauseController.createCommand` and the sealed `PauseApplicationService.execute`), this is a statement switch over the enum with cases for `HELM_RELEASE`/`DEPLOYMENT` and no `default`. Adding a third constant compiles cleanly here and falls straight through — performing **zero validation** for the new mode. This is the one extensibility gap in an otherwise well-guarded design.

Suggest adding `default -> throw new IllegalStateException("Unhandled podDiscoveryMode: " + this);` (or restructuring validation as a per-constant abstract method so the compiler forces each new constant to implement it).

lib/src/main/java/com/scalar/admin/kubernetes/infrastructure/client/KubernetesClientImpl.java
```diff
@@ -62,6 +64,110 @@ public PauseTarget resolvePauseTargetByHelmRelease(String namespace, String helm
     }
   }
 
+  @Override
+  public PauseTarget resolvePauseTargetByDeploymentName(String namespace, String deploymentName, int adminPort)
```
feeblefakie, line 68: **P2 — Deployment mode does no Scalar-product verification and trusts the user-supplied admin port (pause-wrong-workload risk).** This path resolves pods purely from the named Deployment's `spec.selector` and uses the user-supplied `--admin-port`. Unlike the helm path (`resolvePauseTargetByHelmRelease` → `selectPodsRunScalarProduct`), it never confirms the pods carry the `app.kubernetes.io/app` Scalar-product label, and the port isn't derived from a `*-headless` admin Service. Pointing `--deployment-name`/`--admin-port` at the wrong workload would issue Scalar Admin `pause` RPCs against whatever listens on that port.

To be clear, this is operator-driven, not an external-attacker vector — the selector comes from the typed `V1LabelSelector` in the API response, not raw user strings, so there's no injection — and the missing port auto-detection is a deliberate, documented trade-off. But the loss of helm mode's product-verification net is worth a guard: at minimum document the footgun prominently in the README, and ideally run discovered pods through `Product.fromAppLabelValue(...)` and warn/fail-closed when none look like a known Scalar product.

lib/src/test/java/com/scalar/admin/kubernetes/infrastructure/client/KubernetesClientImplTest.java
```diff
@@ -583,6 +589,331 @@ public void select_NormalCase_ShouldReturnPauseTarget() throws Exception {
...
+        assertThatThrownBy(() -> kubernetesClient.getDeployment(namespace, deploymentName))
+            .isInstanceOf(PauserException.class)
+            .hasMessageContaining("Kubernetes readNamespacedDeployment API error with code 404");
+      }
+    }
+  }
+
+  @Nested
+  @DisplayName("extractLabelSelectorFromDeployment()")
+  class BuildLabelSelectorFromDeployment {
```
feeblefakie, line 654: **P2 (test gap) — the `deployment.getMetadata() == null` guard added in this PR is untested.** This `BuildLabelSelectorFromDeployment` class covers valid-selector, no-spec, no-selector, empty-selector, and invalid-operator — but never a deployment with null metadata, so the `"Deployment has no metadata..."` guard (`KubernetesClientImpl.java:93-95`) has zero coverage.

Suggest adding:
```java
@Test
@DisplayName("throws PauserException when deployment has null metadata")
void throwsPauserExceptionWhenDeploymentHasNullMetadata() {
  V1Deployment deployment = new V1Deployment(); // metadata null by default
  assertThatThrownBy(() -> client.extractLabelSelectorFromDeployment(deployment))
      .isInstanceOf(PauserException.class)
      .hasMessageContaining("has no metadata");
}
```

lib/src/test/java/com/scalar/admin/kubernetes/infrastructure/client/KubernetesClientImplTest.java
```diff
@@ -583,6 +589,331 @@ public void select_NormalCase_ShouldReturnPauseTarget() throws Exception {
...
+        String namespace = "default";
+        String deploymentName = "scalardb-cluster";
+        int adminPort = 60054;
+
+        KubernetesClientImpl kubernetesClient =
+            new KubernetesClientImpl(coreV1Api, appsV1Api);
+
+        // Act
+        PauseTarget pauseTarget =
+            kubernetesClient.resolvePauseTargetByDeploymentName(namespace, deploymentName, adminPort);
```
feeblefakie, line 929: **P2 (test gap) — the label selector pass-through is masked by loose `any()` matchers.** This happy-path test stubs `listNamespacedPod(any(), …)`, so the selector extracted from the deployment and threaded into `listNamespacedPod` (the 6th positional arg, `KubernetesClientImpl.java:148-149`) is never asserted. A bug producing the wrong selector — or passing it to the wrong positional parameter — would pass silently. (`adminPort` pass-through is verified with exact `eq()` elsewhere; namespace + selector wiring is not.)

Suggest an `ArgumentCaptor<String>` (or exact-arg `verify`) on `coreV1Api.listNamespacedPod`, asserting `arg[0] == namespace` and `arg[5] ==` the expected selector string the mocked deployment yields. While here, a positive `matchExpressions` selector test (valid `In`/`NotIn`/`Exists`) would also be worth adding — only the invalid-operator path is covered today.

feeblefakie (commented):
## Code review — P2 findings (deployment mode)

Really clean extension of the PR #46 onion architecture — sealed `PauseCommand` variant, compiler-checked dispatch, de-duplicated `executePause(...)`, backward-compatible default. **No P1 blockers.** A few P2s worth addressing before merge are left as inline comments below.

One P2 that can't be left inline because it's outside this PR's diff:

**P2 — `PauseTarget`'s `adminPort` invariant is weaker than the rest of the domain.** `PauseTarget`'s compact constructor (`lib/.../domain/model/pause/PauseTarget.java:45-47`) checks only `adminPort < 1`, while `PauseByDeploymentNameCommand` and `PodDiscoveryMode.validate()` both enforce `[1, 65535]`. Since `PauseTarget` is the aggregate root, any direct construction (test helpers, future callers bypassing the command layer) can silently produce `adminPort = 70000`. Suggest tightening it to `adminPort < 1 || adminPort > 65535` with a shared message constant.

_(Several earlier-round issues are confirmed fixed in the current tree: `getMetadata()` null guard, no double-wrapping of `PauserException` with `isSameAs` tests, `[1,65535]` port range, and the README RBAC `Role`/`RoleBinding`.)_

lib/src/main/java/com/scalar/admin/kubernetes/domain/model/pause/PauseByDeploymentNameCommand.java
monkey-mas: > P2 — Validation rules are duplicated across layers with divergent error messages.

Got it! Consolidated validation so that `Command` compact constructors are the single source of truth for value rules: https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/47/commits/0d51e8930424aaef51aa27b158d5a053e4fd3458
- Removed duplicated `pauseDuration` checks from `PauseRequest` and `PauseService`, and the duplicated  `adminPort` range check from `PodDiscoveryMode.validate()`.
- Shared message constants are now defined on the `PauseCommand` interface.

lib/src/main/java/com/scalar/admin/kubernetes/domain/model/pause/PodDiscoveryMode.java
monkey-mas: Yep, converted the statement switch to a Void switch expression, which gives compile-time exhaustiveness  enforcement. Adding a third `PodDiscoveryMode` constant without a corresponding case will now cause a compile error. This is stronger than a default throw since it catches the gap at compile time rather than runtime.
fix: https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/47/commits/c44bea55f6493ecea090774e336907188487accd

lib/src/main/java/com/scalar/admin/kubernetes/infrastructure/client/KubernetesClientImpl.java
monkey-mas: Thanks for the feedback!

Actually, this is a deliberate design trade-off. Deployment mode exists specifically to support Helm-free  deployments, so requiring the `app.kubernetes.io/app` Scalar-product label (which is set by Scalar Helm Charts) would defeat the purpose of the mode.

Like you suggested, I've added a note in the `README` documenting that deployment mode does not perform automatic product verification or admin port auto-detection, and that the operator is responsible for specifying the correct `--deployment-name` and `--admin-port`.
https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/47/commits/5092c03ed8a30b422d53a559bbb605c642d4c1c7

lib/src/test/java/com/scalar/admin/kubernetes/infrastructure/client/KubernetesClientImplTest.java
monkey-mas: Fixed! https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/47/commits/a757909e39f37181d57151cc3bd41c8c04d8eaf3

lib/src/test/java/com/scalar/admin/kubernetes/infrastructure/client/KubernetesClientImplTest.java
monkey-mas: Done!
https://github.com/scalar-labs/scalar-admin-for-kubernetes/pull/47/commits/797a09fddcc77e0af277b2dec59b42a2a61e2980

brfrn169 (approved):
LGTM! Thank you!

kota2and3kan (approved):
LGTM! Thank you!

feeblefakie (approved):
LGTM! Thank you!

